PR #18, opened 2015-10-22
Sub-monthly eddy computations now supported (this time correctly)

spencerahill:
Until I deleted the faulty logic within the last week or so, aospy supported calculations using eddy (=difference between full values and time-averaged values) data.  However, the logic was conceptually flawed: the time-average was taken as averaging over the entire time domain of input data, and eddies treated as deviations thereto.  But in climate science, we are interested in eddies as variability on sub-monthly timescales.  Therefore, the proper way to compute them is as the deviation from the time mean _within each month_.

The new code supports this, for both regional and lat-lon output data, by simply prepending 'eddy' to the `dtype_out_time` string, e.g. `eddy.av` for lat-lon time-averages using eddy data, or `eddy.reg.ts` for a time-series of region-averages using eddy data.


spencerahill:
Woops, meant for this to merge into xray, not develop.

Issue #14, opened 2015-10-14
Use coordinates directly from xarray.DataArrays, not from parent Model


spencerahill:
Currently, each `Model` object is instantiated with paths to various netCDF files which are used to supply basic information about its grid: latitude, longitude, etc.  Some physical calculations require this grid information in addition to physical variables themselves.  However, the values of these grid properties are not necessarily unique to a `Model`: simulations run in the same model may be output to different grids (horizontal or vertical), or changes in software may make, e.g. the latitude values off on the 5th digit, causing xray to throw an error (the last example comes from @spencerkclark ).

The netCDF files from which data is loaded all contain the coordinate data locally already, and so do the xray.Dataset and/or DataArray objects derived from those files.  So there is no reason to always pull from the Model's grid attributes -- they should be pulled from the local variable itself.

As a starting point, `Calc` should just grab the needed coordinate/grid values from the 1st variable that it loads in that has those values.  Pressure values and vertical pressure thicknesses will likely need to be handled as their own unique case, which is what's done already anyways.

spencerahill:
Now that we have migrated aospy to being largely xarray-based, this has been mostly resolved, in that the functions are now passed `xarray.DataArrays` by default, which have the grid information embedded.  So it's up to the user to write functions that extract these embedded coordinate arrays as needed.

Some functions and variables remain, however, that aren't defined in their own netCDF files but reasonably could be needed by themselves.  Currently these are still pulled from the model.  This aspect still has room for improvement/refactoring.

spencerahill:
One potential solution, although it could be heavy-handed, is to copy all of the grid fields into each DataArray when they are loaded from disk.  So, for example, a t_surf DataArray, which is defined only in lat, lon, and time, would also get `pfull`, `phalf`, `ak`, `bk`, `zsurf`, `land_mask`, etc. added as coordinates.  That way the function can rely on these being available from the DataArrays that get passed in.  These are all sufficiently small relative to the typical actual data arrays of interest that the added computational expense isn't an issue.

However, at least the last two aren't coordinates but data arrays themselves, meaning that we would need to pass in Datasets, not DataArrays, with them included among the variables.  So still need to think this one through.

spencerkclark:
[...]

PR #23, opened 2015-10-26
Towards resolving issue #14

spencerkclark:
Here I make the changes I suggested in my comment in #14.  In the process I realized that it might be important to make sure that the names of coordinates (within the DataArrays we pass around) should match the defined internal names within aospy; this currently is not done in either `calc.py` or `model.py`.  For instance in the xray branch version of `model.py` a diff is done on coordinates named `'lonb'` and `'latb'`; these should instead be the internal names `'lon_bounds'` and `'lat_bounds'` respectively (not the names of the coordinates that were in the file that was read in).  I now try to rename coordinates upon reading in the files so that this is no longer an issue. 

Incidentally I also think we no longer should compute surface area within `calc.py`.  See my comment in `_add_grid_attributes` within `calc.py`.  I think the issue I was having should be resolved by me creating a new Model object, rather than having the existing Model try to accommodate the mis-aligned coordinates.  That said, we should make things very clear via warnings or exceptions if coordinates in Runs and Models are indeed mis-aligned so the user can take the requisite steps to fix things. 

The reason for my change of heart on this is that while it is fairly easy to re-compute surface area based on a new set of `lat` and `lon` coordinates, computing something like `zsurf` or `land_mask` on a slightly different grid is not as trivial.  For that reason I feel like we'd never really satisfactorily be able to support slightly different grids (e.g. off in the 5th digit) for a single Model; it should really be a new model unto itself.  While verbally I might say the model used for each of these cases was the same, for computational purposes in aospy the models are not.

That said, I still think it is valuable to do this (a) to check that the Model and Run grids are aligned and (b) for the convenience of having the peripheral grid information built into each DataArray.


spencerahill:
Thanks a lot for all your hard work on this.  I will take a closer look soon.

aospy/calc.py
```diff
@@ -8,14 +8,13 @@
 
 import numpy as np
 import xray
-
```
spencerahill, line 11: I put builtin imports first, followed by a space, followed by other package imports, followed by a space, followed by relative imports within the package.  So this space should remain.

aospy/calc.py
```diff
@@ -8,14 +8,13 @@
...
 from . import Constant, Var
 from .io import (_data_in_label, _data_out_label, _ens_label, _yr_label, dmget,
                  data_in_name_gfdl)
 from .timedate import TimeManager, _get_time
 from .utils import (get_parent_attr, level_thickness, apply_time_offset,
                     monthly_mean_ts, monthly_mean_at_each_ind,
-                    pfull_from_ps, dp_from_ps, int_dp_g)
+                    pfull_from_ps, dp_from_ps, int_dp_g, to_pascal)
```
spencerahill: `to_pascal` is never used and thus shouldn't be imported

aospy/calc.py
```diff
@@ -254,7 +253,7 @@ def _get_input_data_paths_gfdl_dir_struct(self, name, data_in_direc,
                                                   self.intvl_out,
                                                   self.data_in_start_date[n].year,
                                                   self.data_in_dur[n]))
-                 for year in range(start_year, end_year + 1)]
```
spencerahill: Why is the `+1` removed?  Seems like without it you won't get the `end_year` included.

aospy/model.py
```diff
@@ -74,9 +74,9 @@ def _get_grid_files(self):
         datasets = []
         for path in self.grid_file_paths:
             try:
-                ds = xray.open_dataset(path)
```
spencerahill, line 77: What's the motivation behind this?

aospy/calc.py
```diff
@@ -322,6 +321,53 @@ def _to_desired_dates(self, arr):
...
+
+        for name_int, names_ext in grid_attrs.items():
+            ds_coord_name = set(names_ext).intersection(set(ds.coords))
+            if ds_coord_name:
+                # Check if coord is in dataset already.
+                # If it is, then rename it so that it has 
+                # the correct internal name.
+                ds = ds.rename({list(ds_coord_name)[0] : name_int})
+                ds = ds.set_coords(name_int)
+                if not self._strip_undefined_coords(ds[name_int]).equals(getattr(self.model[n], name_int)):
```
spencerahill: Lines shouldn't exceed 79 characters.

aospy/model.py
```diff
@@ -89,15 +89,38 @@ def _get_grid_attr(self, grid_objs, attr_name):
                 pass
         return
 
+    def _rename_coords(self, ds):
+        """Renames all coordinates within a Dataset or DataArray so that they match 
+        the internal names.
+        """
+        primary_attrs = {
+            'lat':         ('lat', 'latitude', 'LATITUDE', 'y', 'yto'),
```
spencerahill: Let's use the package constants for the keys here, e.g. LAT_STR, LON_STR, etc.  (And anywhere else)

aospy/calc.py
```diff
@@ -322,6 +321,53 @@ def _to_desired_dates(self, arr):
...
+                    print("Warning: Model coordinates for '{}' do not match those in Run".format(name_int))
```
spencerahill: Let's use the builtin `warnings` module (which you introduced me to!) here rather than just printing.  (At some point, I'd like to start using the logging module too to more methodically and systematically log runtime status, metadata of calculations, etc.)

aospy/model.py
```diff
@@ -89,15 +89,38 @@ def _get_grid_attr(self, grid_objs, attr_name):
...
+            'lat_bounds':  ('latb', 'lat_bnds', 'lat_bounds'),
+            'lon':         ('lon', 'longitude', 'LONGITUDE', 'x', 'xto'),
+            'lon_bounds':  ('lonb', 'lon_bnds', 'lon_bounds'),
+            'level':       ('level', 'lev', 'plev'),
+            'phalf':       ('phalf',),
+            'pfull':       ('pfull',)
+            }
+        if isinstance(ds, xray.Dataset) or isinstance(ds, xray.DataArray):
+            for name_int, names_ext in primary_attrs.items():
+                ds_coord_name = set(names_ext).intersection(set(ds.coords))
```
spencerahill, line 111: Nice usage of sets here!

spencerahill:
This looks good overall.  I have made inline comments where there are some concerns/problems.  I'd like to resolve those before making the merge.  (Regarding the style-related ones, I realize that they may seem pedantic, but I'm a big proponent of keeping the style consistent, which includes, as much as possible, being PEP8 compliant.)

I agree that forcing names to be the same is the way to go -- that was the motivation behind creating the `_STR` constants at the package level.  So we should use those throughout the code.  More robust in case we needed to change them for some reason (not likely in this application, but just good practice to create variables for magic numbers/magic strings rather than typing in their value directly (which obviously I only recently have been adhering to myself)).

C.f. our most recent comments in  #14 , would allowing the grid fields to be defined at the `Run` level resolve your issue with the slightly different lat/lon values?  And would that change how we would handle the changes you've implemented.

Thanks a lot!

aospy/calc.py
spencerkclark: Sorry about this (I didn't mean to include this in the PR) -- this is leftover from how I defined the `default_date_range` in my personal Run objects; I would put the end date as January 1st the following year (e.g. if I wanted years [2000, 2004] inclusive I would put my date range as ("2000-01-01", "2005-01-01")).  

As you point out though, if one were to specify a non-yearly interval (like "2000-01-01" to "2004-02-01"), then this logic would fail to retrieve the necessary data for January 2004 (not to mention it isn't consistent with the inclusive-on-both-ends logic in `_get_time()` within timedate.py).  For this reason I will stop being stubborn and fix my Run objects to be consistent with the current aospy conventions.

aospy/model.py
spencerkclark, line 77: Without the `decode_cf=False` argument this line throws an exception in idealized cases whose time units are specified as "days since 0000-00-00 00:00:00", because no library in python supports dates with year 0.

I think we might be able to relax this somewhat by setting `decode_times` to `False` instead.  I'll give it a try.  Do you see any harm in adding the `decode_times` or `decode_cf` arguments here?  My impression is that it doesn't impact the calculations, but safeguards against some mysterious errors.

spencerkclark:
No worries!  I'll take care of the style issues and update the PR later today.  Yes, allowing grid fields (in particular `land_mask` and `zsurf`) to be defined at the Run level would effectively close #14 I think.

spencerahill:
Great, thanks.  Also the commits that I have pushed since you made this (unrelated to this issue) made automatic merging not work, so if its not too much trouble please pull those first.

spencerkclark:
I made the changes and it should be ready to merge automatically now -- I can't view the issues found by QuantifiedCode due to a 404 error on their end.  What kinds of issues does it look for exactly?

spencerkclark:
Actually hold up on merging just yet -- I think I have a cleaner solution than using the `_strip_undefined_coords` method.  The reason I need to use that is that if a grid attribute is not found either in the Model or the Run then it adds a scalar coordinate with value `None` with the name of the attribute.  This would cause problems for the idealized cases, (which don't have a `'level'` attribute or `'land_mask'` etc.), because these undefined scalar coordinates would get carried everywhere.  When you tried to save (or compare to coordinates from the Model object) this would cause exceptions (or coordinate DataArrays that should be considered equal declared not equal due to the presence of the scalar coordinates).

I think a cleaner solution here is to just include some logic to `_add_grid_attributes` along the lines of "if the grid attribute is not in the Run or Model object then throw a Warning, but don't add it to the Dataset."

aospy/model.py
spencerahill, line 77: Ok, thanks.  This is fine with me.

spencerahill:
QuantifiedCode is a startup that has really cool code analysis and visualization tools.  The site has been a little buggy though...I think they are still working out the kinks.  I just invited you as an admin to aospy on their end, let me know if that doesn't solve the problem.  In this case, one thing they are flagging is that you use `isinstance(...) or isinstance(...)`, where the idiom would be to use a single `isinstance` with a tuple of types as the second argument.  If this gets to be a nuisance, we can disable it.  Also, FYI the Travis CI check is totally useless, becase I never set it up properly.  Will do that someday.

(I heard about QuantifiedCode on this podcast, which I really like and you might too: http://talkpython.fm/ )

spencerkclark:
Nice -- I think it was something on my end.  I opened the link in another browser and things worked out fine.  QuantifiedCode looks pretty cool.  I went ahead and addressed all the issues it was finding with my PR (it helps me learn a few things in the process as well; I never really thought about static methods in python before now). 

Great podcast recommendation as well.  Thanks!

spencerahill:
Awesome, thanks Spencer!  This is great stuff.

Issue #42, opened 2015-11-12
Create temporary end-to-end suite of test cases


spencerkclark:
In lieu of a robust set of unit tests (see #22), which will take time to develop, we should have a set of aospy objects (and associated computations) that span the current horizon of the package's uses.  Ideally we'd be able to run a script that would attempt to preform all the specified computations on the set of objects;  if it ran successfully we'd at least have some confidence that it was doing what we wanted (or we'd be able to see what tripped it up).  

The details are still a bit up in the air (we'd have to figure out how to get things to run with one click), but for now I suggest we create a Project that contains Runs from some of the Models we have worked with:
- [x] AM2
- [x] AM3
- [x] HiRAM
- [x] Idealized Moist
- [x] CMIP5 output

We'd then have to come up with a set of computations to do on each of them.  ~~To start out I think these can just be computations that aospy can do without a call to an external `aospy_user/calcs/` function (although I guess there is no reason we couldn't have some dummy functions within this pseudo-aospy_user environment).~~ (This was addressed in #61, which now does allow for calls to external calculation functions located within `test_objs/calcs.py`).  The most basic computations would be (feel free to add more as they come up):
- Annual mean, and time series
- Seasonal mean, and time series
- Monthly mean, and time series
- Vertical integral over interpolated pressure levels
- Vertical integral over sigma pressure surfaces
- Regional average, and time series (with no land mask)
- Regional average, and time series (with land mask)
- Regional average, and time series (with ocean mask)

It would also be great if we could have some reference computations to compare to for these calculations, but that too might be a goal for down the road.  Initially it can just answer the question "does it run or not?"

spencerahill:
How do you feel about this in light of #61?

spencerkclark:
With regard to potentially closing this issue?  Should we wait until we have more coverage, or do you think #61 is good enough for now?  

On the bright side, with #61 it shouldn't be too hard to check a lot of these things off the list.  I can do my part in adding some more tests on AM2 and idealized model output, but I don't have any AM3, HiRAM, or CMIP5 runs handy.  In principle, though, I could just lift some from your object library.

spencerahill:
Ya, just wanted to see if this still has merits, which it does.  I agree, it would be useful (and low-hanging fruit) to copy the other model objects here, since often in the past a change would break things for one model and not the other.

(Doing these does _not_ obviate the need for more unit testing everywhere, of course.)

spencerkclark:
[...]

PR #43, opened 2015-11-13
Added test cases

spencerkclark:
I made a start on adding a test setup here -- feel free to comment on how I've done it.  I tried to keep things minimal.  I basically just added example Runs for AM2 and the idealized moist model and included a basic main script (currently set to compute the mean of the variable `'olr'` in the idealized model).

We still need to determine how to best run multiple tests with one script, and also need to have a way to change the output directory easily for the project (right now it is set to my work directory, so you won't have permission to write to that directory and save calculations done in main.py).


spencerkclark:
No rush to merge this by the way -- I'm more or less posting this here to generate discussion.

spencerahill:
Thanks for starting on this, which addresses #42.  But this needs to be substantially pruned down.  

I don't think we necessarily need to replicate the whole package structure of `aospy-obj-lib`.  But even if we do, we should have only a few bare bones functions and objects.  I'll comment in-line further.

aospy/region.py
```diff
@@ -1,6 +1,6 @@
 """region.py: Region class and region_inst()."""
```
spencerahill, line 1: These changes are unrelated to this issue.

aospy/test/aospy_test_objs/aospy_user/calcs/numerics.py
```diff
@@ -0,0 +1,182 @@
+"""Finite differencing and other numerical methods."""
```
spencerahill: This whole module should be omitted.

aospy/test/aospy_test_objs/aospy_user/projs/.#am2_2009_yim
```diff
@@ -0,0 +1 @@
+skc@public1.24097:1435446444
```
spencerahill: Remove

aospy/test/aospy_test_objs/aospy_user/regions/__init__.py
```diff
@@ -0,0 +1,355 @@
+from aospy.region import Region
```
spencerahill: We should have only a few regions.  One global, one that's a simple lat-lon box, one that has multiple lat-lon bounds, one with land mask, and one without land mask.

aospy/test/aospy_test_objs/aospy_user/sphere_harm.py
```diff
@@ -0,0 +1,62 @@
+"""Interface with windspharm/spharm for spherical harmonics analyses."""
```
spencerahill: This module should be omitted.

aospy/test/aospy_test_objs/aospy_user/variables/__init__.py
```diff
@@ -0,0 +1,967 @@
+"""Collection of aospy.Var objects for use in my research."""
```
spencerahill: There should be only a few variables.  One 2D, one 3D, one at phalf, one at pfull, one that's derived from one or more others, etc.

aospy/test/aospy_test_objs/main.py
```diff
@@ -0,0 +1,213 @@
+#! /usr/bin/env python
```
spencerahill: I don't think we should include all the parsing logic from main.  Can we just create the `Calc` objects and then run them?

spencerahill:
Let me know if any of the comments are unclear.  Ultimately, I think its important (here and elsewhere) to add the minimal amount of code to perform the desired task.  Even if its a temporary fix -- temporary code has a way of sticking around much longer than expected.

Re: the package structure, since we will only have a few of each object type (Model, Run, Var, etc.), I don't think we can just create modules for them at the `aospy_test_objs` level.  Also, let's omit the aospy part and name it `test_objs`.  So e.g. `test_objs/models.py`, `test_objs.runs.py`, etc.

spencerahill:
@spencerkclark , does that all make sense/seem reasonable to you?  Definitely up for further discussion.

> Re: the package structure, since we will only have a few of each object type (Model, Run, Var, etc.), I don't think we can just create modules for them at the aospy_test_objs level. Also, let's omit the aospy part and name it test_objs. So e.g. test_objs/models.py, test_objs.runs.py, etc.

Sorry, this was a typo...the don't should be omitted: I **do** think we can just create modules at this level.

spencerkclark:
Yes, I'm fully on board with pruning things down significantly; sorry for not getting back to you earlier on this (thanks for correcting the typo -- I was a little confused there). 

I'll try and get to this tomorrow if I have time.  Sorry for being quite sloppy in the first iteration (I didn't want to overcommit time to a setup that we wouldn't ultimately use).  Your comments are very helpful.

spencerahill:
Thanks, Spencer.  No rush at all -- just wanted to make sure this didn't fall through the cracks (in that vein, I'll be returning to the AM3 2D bounds issue later tonight).  And I totally hear you re: doing it quick-and-dirty on the first pass.

spencerkclark:
The latest commits cut down a lot on unnecessary code.  Please let me know if there's anything else you'd like me to remove.  I think I might need your help in getting rid of the parsing logic in main.py.  For the simple calculation that is outlined there now, could you give me an example of how to create and execute it without the parsing logic?  Thanks

spencerahill:
Hey I'm really sorry this got forgotten.  What is the status of this nowadays, given some seemingly similar developments you've made in aospy-db?

spencerkclark:
Thanks for the reminder -- yes the progress in aospy-db should make this much easier.  I'll try and get to this next week if I get a chance.

spencerahill:
Closing in favor of #61

PR #56, opened 2016-01-29
Relax np.any() condition in populating model grid attributes

spencerkclark:
I forget, was there a particular reason we used `np.any()` here?  In certain configurations of the idealized moist model, `pk` is all zeros;  this causes `np.any()` to return `False`, and subsequently `pk` is dropped from the model object, which causes problems downstream.  I think a reasonable fix is to just check the attribute has no object associated with it (i.e. just relax the conditional statement a bit).


spencerahill:
Ya this is fine.  Thanks!

PR #57, opened 2016-01-31
Testing timedate.py

spencerkclark:
Because it was a somewhat low-hanging fruit and I wanted to get some practice writing tests, I added some tests for some of the TimeManager methods within `timedate.py`. 
- In doing so I noticed that `TimeManager.month_indices` would return `[12, 13, 14]` when provided `'djf'` as an argument.  I guess this didn't cause any issues downstream, but out of safety I modified the method a bit.  
- I also noticed that if a string was provided, as long as the first letter was in `'jfmamjjasondjfmamjjasond'`, then it would always return something (even if the input was ambiguous, e.g. `'j'`, or meaningless, e.g. `'jq'`, or a nonconsecutive set of months, e.g. `'jjam'`).  This was partly because `str.find(arg)` returns the index within `str` of the start of `arg` if `arg` is in the `str` and `-1` if `arg` is not in `str`.  I added some logic to raise a `RuntimeError` if an input could not be parsed properly. 

Eventually I'll get to testing `TimeManager.create_time_array()` and `timedate._get_time()`, but I wanted to address the small issues with `month_indices` first.  Let me know if you have any comments or issues with how I set up the tests or modified `month_indices`.  Thanks!

### DIFF
--- a/aospy/test/test_timedate.py
+++ b/aospy/test/test_timedate.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-"""Test suite for aospy.utils module."""
+"""Test suite for aospy.timedate module."""
 import sys
 import unittest
 
@@ -12,7 +12,6 @@ from aospy.timedate import TimeManager
 
 class AospyTimeManagerTestCase(unittest.TestCase):
     def setUp(self):
-#        self.tm = TimeManager()
         pass
 
     def tearDown(self):
--- a/aospy/test/test_timedate.py
+++ b/aospy/test/test_timedate.py
@@ -5,9 +5,11 @@ import unittest
 
 import numpy as np
 import xarray as xr
+import pandas as pd
 from datetime import datetime
 
 from aospy.timedate import TimeManager
+from aospy import TIME_STR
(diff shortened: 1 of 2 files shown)


spencerahill:
This is great.  Catching edge cases and small logical errors like these is exactly why we need tests.

For your reference, I have added a fair amount of tests to infinite diff: https://github.com/spencerahill/infinite-diff/blob/develop/infinite_diff/test/test_finite_diff.py.  I kept creating bugs and not realizing it, to the point where I finally gave in and made a test suite.  

Doing so has made me realize that, for all its benefits, xarray actually makes bugs harder to find, because even if your logic is flawed it will usually not raise an error, since it automatically broadcasts dimensions, etc.  All the more reason for better testing as we move forward.

spencerahill:
One testing tip: make sure that each test you write can fail.  I have made tests before that were useless, because their own logic made them not actually test what I wanted.  So be sure to run the test on the code with the bug, see the test fails, fix the bug, and see the test pass.  You might already be doing this :)

PR #61, opened 2016-02-04
End-to-end test module

spencerkclark:
This is a start of a minimal end-to-end testing module (as discussed in #42).  Currently I've just included a test of one project, one model, one run, and one model-native variable.  The structure is based off comments in [pull request 43](https://github.com/spencerahill/aospy/pull/43) and further developments in spencerkclark/aospy-db#2.

There is ample room to make this more comprehensive, but I want to get the basic structure down first.  (This is meant to replace pull request 43; rather than modify the existing one, I thought it might be cleaner to start fresh).

Please let me know if you have any comments or suggestions.  Thanks!

### DIFF
--- a/aospy/test/test_objs/models.py
+++ b/aospy/test/test_objs/models.py
@@ -1,0 +1,19 @@
+from aospy.model import Model
+import runs
+
+
+am2 = Model(
+    name='am2',
+    grid_file_paths=(
+        ('/archive/Yi.Ming/sm2.1_fixed/SM2.1U_Control-1860_lm2_aie_rerun6.YIM/'
+         'pp/atmos/atmos.static.nc'),
+        ('/archive/Yi.Ming/sm2.1_fixed/SM2.1U_Control-1860_lm2_aie_rerun6.YIM/'
+         'pp/atmos_level/atmos_level.static.nc'),
+        ('/archive/Yi.Ming/sm2.1_fixed/SM2.1U_Control-1860_lm2_aie_rerun6.YIM/'
+         'pp/atmos_level/ts/monthly/5yr/atmos_level.011601-012012.vcomp.nc'),
+        ('/archive/Yi.Ming/sm2.1_fixed/SM2.1U_Control-1860_lm2_aie_rerun6.YIM/'
+         'pp/atmos/ts/monthly/100yr/atmos.000101-010012.temp.nc')
+    ),
+    runs=[runs.test_am2],
+    default_runs=[runs.test_am2]
+)
--- a/aospy/test/test_roundtrip.py
+++ b/aospy/test/test_roundtrip.py
@@ -1,14 +1,10 @@
 #!/usr/bin/env python
 """
-02-04-2016 SKC: Set of end-to-end tests of aospy.  Currently set up to test for
+Set of end-to-end tests of aospy.  Currently set up to test for
 runtime errors, but could be extended to test the numerical results.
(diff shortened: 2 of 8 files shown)


spencerahill:
From first glance this looks great.  I'll take a closer look soon.  Thanks!

aospy/test/test_roundtrip.py
```diff
@@ -0,0 +1,48 @@
...
+                                 dtype_out_time='av',
+                                 dtype_out_vert=False,
+                                 level=False)
+        calc = Calc(calc_int)
+        calc.compute()
+
+
+if __name__ == '__main__':
+    suite = unittest.TestLoader().loadTestsFromTestCase(TestAospy)
+    unittest.TextTestRunner(verbosity=2).run(suite)
```
spencerahill: What's the motivation behind these calls?  I think the idiom for the **main** block of test modules is just `unittest.main()` (see e.g. [test_utils.py](https://github.com/spencerahill/aospy/blob/develop/aospy/test/test_utils.py#L123)).

aospy/test/test_objs/models.py
```diff
@@ -0,0 +1,19 @@
+from aospy.model import Model
+import runs
+
+
+am2 = Model(
```
spencerahill, line 5: Here and elsewhere that has GFDL-specific files, we might want to put in some try/except guards such that these don't cause the overall aospy testing suite to crash from non-GFDL machines (thinking about this led me to open #62).

For the tests themselves, we can use unittest's [skip decorators](https://docs.python.org/2/library/unittest.html#unittest.skip).

aospy/test/test_roundtrip.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+"""
+02-04-2016 SKC: Set of end-to-end tests of aospy.  Currently set up to test for
```
spencerahill: You can omit the date and initials...I know I started us down that path a while back, but now I've realized that Github's blame makes this unnecessary (and the more discussion-oriented comments I use to include are much better, I now realize, as Github issues).

spencerahill:
Other than my in-line comments, this looks great and works for me from an ssh.  Out of curiosity I cloned it to my laptop also, and the result is just a failure of test_am2_annual_mean with `RuntimeError: No such file or directory`.  So I think all we need is a `skipIf` decorator, with the boolean being a check if the needed file exists -- probably [os.path.isfile](https://docs.python.org/2/library/os.path.html#os.path.isfile).

This PR, combined with #63, will give me so much more piece of mind moving forward.

aospy/test/test_objs/models.py
spencerkclark, line 5: Let me know if 973d868 addresses this to your liking.  It's not as general as we might like (since the test doesn't just look for the grid files, it also looks for the olr files), but I think it is sufficient to determine whether a test is only possible to be run in the GFDL environment.

aospy/test/test_roundtrip.py
spencerkclark: No real motivation here; this is more or less just carried over from me experimenting with different levels of verbose output from the `unittest` module.  I have no issues making this more consistent with the other test modules already defined in `aospy`.

spencerahill:
Yes, this works perfectly -- skips that test on my local machine, and test runs and passes at GFDL.

Thanks a lot!

PR #64, opened 2016-02-05
Add more coverage to AM2 end-to-end tests

spencerkclark:
Addresses the test items listed in #42 for the AM2 model.  Also makes code to define individual tests more concise.


spencerahill:
This is great, thanks yet again!

PR #66, opened 2016-02-09
TEST Add test cases for AM3, HiRAM, and CMIP5 output

spencerkclark:
Addresses test operations suggested in #42 for AM3, HiRAM, and CMIP5 output.  Currently all CMIP5 tests fail, because no grid files are provided in the model objects.  @spencerahill when was the last time you used `aospy` on CMIP5 output?  Would you expect things to work, or might some modifications need to be made?  Might a simple fix be just to add some grid files?

### DIFF
--- a/aospy/test/test_roundtrip.py
+++ b/aospy/test/test_roundtrip.py
@@ -246,14 +246,14 @@ class TestHiRAM(TestAM2):
                                  'dtype_in_time': 'ts',
                                  'level': False}
 
-    @unittest.skip(('could not immediately find a case with temperature'
-                    ' output on sigma levels in CMIP5 archive'))
+    @unittest.skip('no HiRAM output readily at hand exists on sigma levels')
     def test_vert_int_sigma(self):
         pass
 
 
 # cesm1_cam5 has no grid files so just use hiram's to test to skip
 @unittest.skipIf(not model_files_exist(hiram), skip_message)
+@unittest.expectedFailure
 class TestCMIP5(TestAM2):
     def setUp(self):
         self.olr_test_params = {'proj': aospy_test,
@@ -276,8 +276,8 @@ class TestCMIP5(TestAM2):
                                  'dtype_in_time': 'ts',
                                  'level': False}
 
-    @unittest.skip(('could not immediately find a case with temperature'
-                    ' output on sigma levels in CMIP5 archive'))
+    @unittest.skip(('This repo does not contain any output'
+                    ' on sigma levels.'))
     def test_vert_int_sigma(self):
         pass
 
--- a/aospy/test/test_roundtrip.py
+++ b/aospy/test/test_roundtrip.py
@@ -8,21 +8,35 @@ from os.path import isfile
 
 from aospy.calc import Calc, CalcInterface
 from test_objs.projects import aospy_test
-from test_objs.models import am2, idealized_moist
-from test_objs.runs import test_am2, test_idealized_moist
+from test_objs.models import am2, idealized_moist, am3, hiram, cesm1_cam5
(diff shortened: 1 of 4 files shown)


spencerahill:
Hmm I definitely got the CMIP5 models working within aospy (summer 2015), but I don't see how they worked without grid files, nor do I recall how I handled the grid files.  I will have to look into that.  Their mechanics are otherwise different from our other data, in that they are a formal GFDL 'repo', which Calc has special methods for handling.  This allowed me to create Run objects once which multiple models could use (since in CMIP everybody is running the same simulations).

For the time being, maybe the best thing is to mark the CMIP5 tests with the [expected failure](https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures) decorator.  I'll create an issue in my aospy-obj-lib to keep track of that.

aospy/test/test_roundtrip.py
```diff
@@ -184,5 +198,89 @@ def test_monthly_ts(self):
...
+                                 'model': cesm1_cam5,
+                                 'run': test_amip,
+                                 'var': temp,
+                                 'date_range': ('1979-01-01',
+                                                '2008-12-31'),
+                                 'intvl_in': 'monthly',
+                                 'dtype_in_time': 'ts',
+                                 'level': False}
+
+    @unittest.skip(('could not immediately find a case with temperature'
```
spencerahill: I think the CMIP5 data is all interpolated to regular pressure levels, so we'll just have to omit this test.

aospy/test/test_roundtrip.py
```diff
@@ -184,5 +198,89 @@ def test_monthly_ts(self):
...
+                                 'model': hiram,
+                                 'run': test_hiram,
+                                 'var': temp,
+                                 'date_range': ('1979-01-01',
+                                                '1995-12-31'),
+                                 'intvl_in': 'monthly',
+                                 'dtype_in_time': 'ts',
+                                 'level': False}
+
+    @unittest.skip(('could not immediately find a case with temperature'
```
spencerahill: Message is about CMIP5 but this is the HiRAM class.  Intentional?  Confusing.

aospy/test/test_roundtrip.py
```diff
@@ -184,5 +198,89 @@ def test_monthly_ts(self):
...
+                    ' output on sigma levels in CMIP5 archive'))
+    def test_vert_int_sigma(self):
+        pass
+
+
+# cesm1_cam5 has no grid files so just use hiram's to test to skip
+@unittest.skipIf(not model_files_exist(hiram), skip_message)
```
spencerahill, line 255: Here's where I'd add the expected failure decorator.  I think using both will work -- we want to keep the skip decorator, for testing not on the GFDL machines, i.e. we want to skip if files aren't there and expected failure if files are there.

aospy/test/test_roundtrip.py
spencerkclark: Good catch, not intentional.  Are you aware of any runs where temperature is output on sigma levels in HiRAM?

spencerkclark:
Yeah no worries about the CMIP5 cases; I just wanted to bring that to your attention.  I think using the expected failure decorator is a good solution for now.  I'll address all these things soon.

aospy/test/test_roundtrip.py
spencerahill: No, the HiRAM runs I use were all done by Yi, and I don't think he outputted much (if any) data on model coordinates.

spencerahill:
Awesome, thanks!

spencerkclark:
This should be ready to go!  Let me know if you have any further comments.

spencerahill:
:+1: Thanks!

Issue #51, opened 2015-12-31
Plot profiles of model-native vertical coordinate data to correct levels, not reference levels


spencerahill:
Currently, plotting a profile of data that is defined on model-native vertical coordinates results in the reference `pfull` (or `phalf`) array being used for the pressure value of each level.  Unless the region of interest has a surface pressure that matches that used for the reference array, namely 1013.25 hPa, these pressure values are incorrect for the data being plotted.  For example, I just created some profiles for a region with a surface pressure of 960 hPa, but its lowest level is (incorrectly) plotted to be at 1013.25 hPa.

The plotting routines should therefore use the actual pressure values for that data, which requires knowledge of the surface pressure.

spencerahill:
Closed in #74

PR #67, opened 2016-02-22
Save region-averaged pfull values and then load them

spencerahill:
This is an attempt to address #51.

@spencerkclark, would you mind taking a look at this when you get a chance?  What do you think of this implementation?  It was trickier than I initially thought -- I had to rename each region's `pfull` values by appending the region name, because otherwise the coord has conflicting values within the dataset.  To load the region values, I drop all of the other region pressure values (because each region ends up with every region's pressure array as coords) and overwrite the original `pfull` with the region's actual pressure values.  I'm wondering if there was an easier way...

Also, how does it relate to https://github.com/spencerkclark/aospy-db/issues/1?  Thanks!

(The most recent commit, f595017, is the one with all of these changes.  Note that, in poor practice, a decent amount of cleanup/mods that aren't related to this actual feature add ended up in that same commit.  Sorry about that -- let me know if anything is confusing.  I'm running our test_roundtrip suite right now, and they're passing so far.)

### DIFF
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -785,7 +785,7 @@ class Calc(object):
                            region=False):
         """Load aospy data saved on scratch file system."""
         ds = xr.open_dataset(self.path_scratch[dtype_out_time],
-                               engine='scipy')
+                             engine='scipy')
         if region:
             return ds[region.name]
         return ds[self.name]
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -12,12 +12,13 @@ import xarray as xr
 
 from . import Constant, Var
 from .__config__ import (LAT_STR, LON_STR, LAT_BOUNDS_STR, LON_BOUNDS_STR,
-                         PHALF_STR, PFULL_STR, PLEVEL_STR, TIME_STR, YEAR_STR)
+                         PHALF_STR, PFULL_STR, PLEVEL_STR, TIME_STR, YEAR_STR,
+                         ETA_STR)
 from .io import (_data_in_label, _data_out_label, _ens_label, _yr_label, dmget,
                  data_in_name_gfdl)
 from .timedate import TimeManager, _get_time
 from .utils import (get_parent_attr, apply_time_offset, monthly_mean_ts,
-                    monthly_mean_at_each_ind, pfull_from_ps,
+                    monthly_mean_at_each_ind, pfull_from_ps, to_hpa,
                     to_pfull_from_phalf, dp_from_ps, dp_from_p, int_dp_g)
 
 
@@ -241,8 +242,8 @@ class Calc(object):
             elif os.path.isfile(full):
                 paths.append(full)
             else:
-                print("Warning: specified netCDF file `{}` "
-                      "not found".format(nc))
+                warnings.warn("Warning: specified netCDF file `{}` "
+                              "not found".format(nc))
         # Remove duplicate entries.
         files = list(set(paths))
         files.sort()
@@ -261,7 +262,7 @@ class Calc(object):
         dtype_lbl = self.dtype_in_time
         if self.intvl_in == 'daily':
             domain += '_daily'
-        if self.dtype_in_vert == 'sigma' and name != 'ps':
+        if self.dtype_in_vert == ETA_STR and name != 'ps':
             domain += '_level'
         if self.dtype_in_time == 'inst':
             domain += '_inst'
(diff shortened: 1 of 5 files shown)


aospy/calc.py
```diff
@@ -650,9 +660,36 @@ def region_calcs(self, arr, func, n=0):
             else:
                 method = getattr(reg, func)
                 data_out = method(arr)
-            reg_dat.update({reg.name: data_out})
+                if bool_pfull:
+                    # Don't apply e.g. standard deviation to coordinates.
+                    if func not in ['av', 'ts']:
+                        method = reg.ts
+                    # Convert Pa to hPa
+                    coord = method(pfull) * 1e-2
```
spencerkclark, line 668: Should we be using the `utils.to_hpa()` method here?

spencerkclark:
Good to hear those tests are being put to good use!

> It was trickier than I initially thought -- I had to rename each region's pfull values by appending the region name, because otherwise the coord has conflicting values within the dataset. To load the region values, I drop all of the other region pressure values (because each region ends up with every region's pressure array as coords) and overwrite the original pfull with the region's actual pressure values.

This does seem like a tricky problem -- see my inline comment (below) for my brief thoughts.  It may or may not be a better solution, but I'm just throwing that out there.

Regarding spencerkclark/aospy-db#1, I don't think this should be a big deal.  Currently, anything related to coordinates will not play a role (since we do not track those as aospy Var objects);  essentially the coordinates are just something brought along with the output file, which we index in the database (at least in part) by the variable name stored.  So when we query for a variable (in this case the variable name and the region name), we get the relevant coordinates automatically.

aospy/calc.py
```diff
@@ -785,9 +808,19 @@ def _load_from_scratch(self, dtype_out_time, dtype_out_vert=False,
...
         ds = xr.open_dataset(self.path_scratch[dtype_out_time],
-                               engine='scipy')
+                             engine='scipy')
         if region:
-            return ds[region.name]
+            arr = ds[region.name]
+            # Use region-specific pressure values if available.
+            if self.dtype_in_vert == ETA_STR and not dtype_out_vert:
+                # try:
+                    arr[PFULL_STR].values = arr[region.name +
```
spencerkclark, line 817: I just want to make sure I understand things correctly: are the region-specific pressure coordinates and the regional variables themselves functions of the universal pressure coordinate `PFULL_STR`?  

One way you could avoid this dropping maneuver would be to store the regional variables as functions of the region-specific pressure coordinates (and have the region-specific coordinates be functions of themselves (not the universal `PFULL_STR`)).  Then in passing the region-specific DataArray to your plotting routines all you would need to do is select the region from the Dataset and rename its pressure coordinate to `PFULL_STR`.  If things were structured this way, selecting from the Dataset would automatically take care of dropping the other coordinates.  It would also remove the need to replace coordinate values.

It might be slightly cleaner down here, but perhaps upstream (in initially saving the data) it could complicate things, so you'd have to make that judgement.

spencerahill, line 817: > are the region-specific pressure coordinates and the regional variables themselves functions of the universal pressure coordinate PFULL_STR? 

Yes

> One way you could avoid this dropping maneuver would be to store the regional variables as functions of the region-specific pressure coordinates (and have the region-specific coordinates be functions of themselves (not the universal PFULL_STR)). Then in passing the region-specific DataArray to your plotting routines all you would need to do is select the region from the Dataset and rename its pressure coordinate to PFULL_STR. If things were structured this way, selecting from the Dataset would automatically take care of dropping the other coordinates. It would also remove the need to replace coordinate values.

I like that idea.

> It might be slightly cleaner down here, but perhaps upstream (in initially saving the data) it could complicate things, so you'd have to make that judgement.

Good point.  I think I will still ultimately implement your suggestion though -- I think it's better to have the data saved to disk with each region defined in terms of its own coordinates.  (Maybe also include the reference `pfull` array as an extra coord?)

FYI it will probably be a while (i.e. next week or later) before I'm able to make these changes...trying to finish a manuscript, and its sufficiently functional for now.

aospy/calc.py
spencerahill, line 668: Yes, this was laziness on my part.  The warning message about the conversion was distracting to me, so I just did this.  I will tweak `to_hpa` for better control of the warnings.

spencerahill:
Thanks a lot @spencerkclark !  I am closing this and re-opening it vs. the prior commits that I've now pushed to develop, so that it's cleaner.

PR #68, opened 2016-02-22
Feature/save region pfull

spencerahill:
See #67

### DIFF
--- a/aospy/__config__.py
+++ b/aospy/__config__.py
@@ -3,6 +3,7 @@ import os
 default_colormap = 'RdBu'
 user_path = os.path.join(os.getenv('HOME'), 'aospy_user', 'aospy_user')
 
+ETA_STR = 'sigma'
 LON_STR = 'lon'
 LAT_STR = 'lat'
 LON_BOUNDS_STR = 'lon_bounds'
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -187,9 +187,7 @@ class Calc(object):
 
     def _print_verbose(self, *args):
         """Print diagnostic message."""
-        if not self.verbose:
-            pass
-        else:
+        if self.verbose:
             try:
                 print('{} {}'.format(args[0], args[1]),
                       '({})'.format(time.ctime()))
@@ -509,9 +507,6 @@ class Calc(object):
             if self.dtype_in_vert == ETA_STR:
                 if self.dtype_in_time == 'inst':
                     data = self._correct_gfdl_inst_time(data)
-                    self._pfull_data = self._correct_gfdl_inst_time(
-                        self._pfull_data
-                    )
                 return self._to_desired_dates(data)
             return data
         # Get grid, time, etc. arrays directly from model object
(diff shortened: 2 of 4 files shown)


spencerahill:
Note that implementing this creates an unintended consequence: output from two different simulations, or time ranges, etc. can no longer be easily compared (as in one minus the other, or some other numerical operation), because their coordinate values differ (since ps will differ between them).  So before going through with this, we need to think about how to handle this.

One option would be to revert to the reference `pfull` array in this case, but that takes away the whole purpose!  Physically speaking, I think it would make sense to average the `pfull` values of all of the different pieces of data involved in the operation and use that for the resulting data.

Note that the only place within aospy that these operations are performed is within [plotting.py](https://github.com/spencerahill/aospy/blob/develop/aospy/plotting.py#L569).  But it's nevertheless important to get straightened out.

spencerahill:
Thinking more about it, just averaging the two coordinate arrays and re-gridding both original values to these new averaged coordinates is a good solution for now.

Eventually, we could interpolate one or the other array to the opposite one's coordinates (or both of them to their average).  But even the aforementioned simple averaging is more physically sound than just subtracting the values at the reference pressure levels.

spencerkclark:
Good point, I hadn't thought about what would happen in the case of a comparison between regions from two different runs.  In that case I think the way you did things initially might be the simplest option for the time being.  At least in that case you can take the average easily by just summing the pressure DataArrays and dividing by two (since they will both be indexed by the universal pressure coordinate).  If the pressure DataArrays are indexed by different pressure coordinates (as in my suggested option), this operation isn't as trivial.

Unfortunately outside of a refactor where single regions end up in separate files (which could be disruptive to the current setup), I'm not sure if there's a way of avoiding having to do something messy (either in saving the data initially, loading the data, or plotting it).  I feel like we can clean up one area, but it makes a mess somewhere else.

spencerahill:
> At least in that case you can take the average easily by just summing the pressure DataArrays and dividing by two (since they will both by indexed by the universal pressure coordinate). If the pressure DataArrays are indexed by different pressure coordinates (as in my suggested option), this operation isn't as trivial.

Good point.  So I was accidentally right :)

> I feel like we can clean up one area, but it makes a mess somewhere else.

I agree, no unambiguous best option.  But definitely based on this we'll save the reference coordinate array in addition to the new coordinates, for this and any future coordinate transforms that come up.

spencerahill:
I've gotten this working by, when loading the regional data, retaining the reference pfull array, renaming it to `'pfull_ref'`, renaming the region-specific pfull to `'pfull'`, and finally adding some logic to plotting.py to do the coordinate averaging.

@spencerkclark any further thoughts?  This seems like a decent solution to me, and it was easier to implement than I thought.  

I'm gonna hold off on merging this even after your comments until I get a chance to write some unit tests (which I would like for us to start doing for all new PRs moving forward).

spencerkclark:
I agree, this is a solid solution given the way things are setup.

> I'm gonna hold off on merging this even after your comments until I get a chance to write some unit tests (which I would like for us to start doing for all new PRs moving forward).

:+1: on this as well.  Go ahead and merge when you are ready!

spencerahill:
Awesome, thanks.

Issue #71, opened 2016-03-10
Extend fix for years less than 1678 for non-"days since" time units


spencerkclark:
Currently our workaround for dates with years less than 1678 makes the assumption that the original units of the time dimension in the read-in file were `"days since ..."`.  As part of the workaround we hard-code the units to be `"days since 1900-01-01 00:00:00"` (see https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L417-L418) and then decode the CF properties.  This causes problems for high-frequency output, because while the units of the time dimension in the original file might be `"hours since ..."`, they get changed to `"days since 1900-01-01 00:00:00"`.  This means that xarray interprets the time interval between outputs as days (rather than hours in this example); e.g. 2920 3-hourly outputs get interpreted to be 24 years (rather than one year).  
- A straightforward fix would be to allow the user to specify another argument when setting up a calc to specify the time units; that being said I'm not sure how keen we are in adding more arguments to that interface.  
- It also might be possible to do this automatically (by searching for one of `['hours', 'days']` in the units string of the original file).  I would lean towards this solution, with sufficient testing.

For the time being (since I don't switch back and forth between doing high-frequency and low-frequency analysis too often) it's fine for me to correct this by changing the hard-coding (e.g. change the time units to `"hours since 1900-01-01 00:00:00"`), but I definitely think this is something that needs to be addressed eventually (although hopefully at some point this gets fixed upstream, so we don't have to use a workaround to begin with).

spencerahill:
Good catch.  How about something like this:

``` python
freq = test[v].attrs['units'].split('since')[0]
test[v].attrs['units'] = freq + 'since 1900-01-01 00:00:00'
```

Haven't tested this even informally but seems like it should work.

spencerkclark:
:+1: That looks like a very clean way to go.  I'll give it a try tomorrow.

That's also more or less how it's done in xarray: https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L83-L91

spencerahill:
Nice.  I'm sure the way they do it is more robust, so you can emulate them directly if you'd like.  Your call if its worth the trouble -- as you noted, for us this is ultimately a temporary workaround anyways.

spencerkclark:
@spencerahill by the way, your suggested workaround is good in my case -- the only possible advantage I see in using the xarray method is that it is robust to extraneous whitespace.  Beyond that, I think the two methods basically do the same thing.

[...]

Issue #72, opened 2016-04-27
For 'gfdl_repo' and 'one_dir' data, missing data silently ignored


spencerahill:
If a date range is specified but the data is only found for a subset of that range, computations are performed on the available data without any indication of the incorrect range, _and_ when saved to disk the label incorrectly indicates the full specified range.  

For example, If I ask for 1979-2008 averages on this precipitation data: /archive/pcmdi/repo/CMIP5/output/NSF-DOE-NCAR/CESM1-CAM5/amip/mon/atmos/Amon/r1i1p1/v20120612/pr/pr_Amon_CESM1-CAM5_amip_r1i1p1_197901-200512.nc, the output will be saved as 1979-2008 despite the data stopping at 2005.

Should be straightforward to fix -- just requires adding some checks and raising an exception.

spencerkclark:
Tangentially related, but does this mean you've worked out spencerahill/aospy-obj-lib#10?

spencerahill:
Yes; see update to that issue (with apologies for not much detail).  But I've been performing calculations on CMIP5 `amip` and `amip4K` runs for the past several weeks.

spencerahill:
Also as an aside the data originally referenced in this issue is bad, both for `amip` and `amip4K`: http://www.cesm.ucar.edu/CMIP5/errata/

spencerkclark:
> If a date range is specified but the data is only found for a subset of that range, computations are performed on the available data without any indication of the incorrect range, and when saved to disk the label incorrectly indicates the full specified range.

I recently encountered this for files retrieved under the `'one_dir'` paradigm as well, so we should make sure the fix for this issue we settle on addresses both `'gfdl_repo'` and `'one_dir'`.

spencerkclark:
Closed by #90

PR #73, opened 2016-05-01
BUG: Allow for non "days-since" units for runs with years less than 1678

spencerkclark:
This addresses #71, using the code suggested by @spencerahill.  A test has been added in the end-to-end test suite, `test_sub_daily`, to verify that we can now support any time unit in runs with years less than 1678.  Adding this test required adding new Model and Run objects corresponding with the idealized moist model with full radiation (in contrast to the standard configuration of the idealized moist model, the full radiation version outputs sub-daily data with `'hours since...'` time units rather than `'days since...'` in the standard configuration).

In addition, since the idealized moist model with full radiation does not have an `olr` diagnostic, I've switched the 2D test to be on `t_surf` for all models.


spencerahill:
Thanks Spencer.  I will look this and #72 over in a week or so -- behind on my thesis right now.  But initial glance looks good.

aospy/test/test_roundtrip.py
```diff
@@ -150,33 +151,53 @@ def test_vert_int_sigma(self):
...
                                  dtype_in_vert='sigma',
-                                 **self.temp_test_params)
+                                 **self.three_d_test_params)
+        calc = Calc(calc_int)
+        calc.compute()
+
+    def test_sub_daily(self):
+        """Not tested for now in models other than idealized_moist.
+        and idealized_moist_rad."""
+        if not getattr(self, 'sub_daily_test_params', False):
```
spencerahill, line 161: Can you you use the `unittest.skipif` decorator here instead?  That would be more idiomatic and consistent with the rest of the code.

aospy/test/test_roundtrip.py
```diff
@@ -199,52 +220,104 @@ def test_vert_int_pressure(self):
...
+                                      'model': idealized_moist_rad,
+                                      'run': test_idealized_moist_rad,
+                                      'var': temp,
+                                      'date_range': ('0003-01-01',
+                                                     '0006-12-31'),
+                                      'intvl_in': '3-hourly',
+                                      'dtype_in_time': 'inst',
+                                      'level': False}
+
+    @unittest.skip('not tested in idealized moist rad model')
```
spencerahill: As an aside, the fact that we have all these skips for the idealized moist case means that our overall structure should be improved...there should be separate mix-ins for the tests applicable to all models and for those only applicable to the full physics one.  (That was my mistake)

Doesn't need to be addressed in this PR though, nor is it that important, just thinking out loud

aospy/test/test_roundtrip.py
spencerkclark, line 161: That was my initial thought too, but the simplest way I could think of to check if we wanted to run the sub-daily test for a given model was to see if we defined the `sub_daily_test_params` attribute in the given class.  This requires referencing `self`, which unfortunately I can't reference outside the method body.

Do you think there is a better way to do this using `unittest.skipif` instead?  This perhaps could also be a place where re-thinking the structure could help...but like you say given our commitments at the moment, not a huge priority.

aospy/test/test_roundtrip.py
spencerkclark: Agreed.  In my latest commit, 6af1551, I redefine `TestIdealizedMoistRad` as extending `TestIdealizedMoist` rather than `TestAM2`.  This eliminates the need for re-defining the skipped methods for the second idealized model.  This gets us some of the way there within the scope of this PR.

aospy/test/test_roundtrip.py
```diff
@@ -39,100 +40,100 @@ def model_files_exist(model):
 @unittest.skipIf(not model_files_exist(am2), skip_message)
 class TestAM2(unittest.TestCase):
     def setUp(self):
-        self.olr_test_params = {'proj': aospy_test,
```
spencerahill, line 42: FYI I like starting long, multi-line sequences like this on the next line, e.g.

``` python
self.olr_test_params = {
    'proj': aospy_test,
    ...
}
```

That way you don't have to change the spacing (and it won't show up on a diff) if you change the name.  Don't worry about here, but for future reference :)

aospy/test/test_roundtrip.py
spencerahill, line 161: Hmm...could you maybe get to it through one of the dunder attributes, like `__class__`?  At the same time, because these derive from `unittest.TestCase` and use `setUp`, inheritance stuff can be pretty whacky.

Ignore for now -- surely we'll revamp the test suite sooner or later, and if this is still around at that point we'll revisit.

spencerahill:
Sorry this took awhile...all looks good.  I'm waiting on the tests to finish running, then if all pass I'll merge.

spencerahill:
@spencerkclark How long does running the test suite usually take for you?  I'm at ~25 min and counting, still only ~2/3 through test_roundtrip.  Need to leave the office soon-ish, so might need to kill the ssh session they're running in before they finish!  Definitely should think about this during the big re-factor.

spencerahill:
Ok have to leave now and kill these (50+ min)...will restart the tests once home and then merge this.

spencerkclark:
Yeah...it can take a while depending on how recently the reference files have been accessed on archive.  

> Definitely should think about this during the big re-factor.

I fully agree -- ultimately I think adding another round trip test for a small issue like this is overkill (ideally we'd test this PR with a synthetic dataset).  That said, I hesitate to go the full synthetic dataset / new test suite route in this PR, just because I think down the road during the refactor we'll probably re-write most of these tests anyway.  It will be really nice if we can get to the point where the continuous integration actually tests a significant portion of the code...

Sorry if this kept you at the office longer than you wanted to be there today -- no rush at all to merge this.

aospy/test/test_roundtrip.py
spencerkclark, line 42: Ah, that is a much cleaner way of doing things; thanks for pointing that out.

spencerahill:
OK #68 caused some conflicts so I had to manual merge...in so doing tried to rebase to squash some of the commits and failed miserably.  So the commits got weird but ultimately this is now implemented in  372c6b8

@spencerkclark thanks a lot and please let me know if I messed any of it up in my attempts to be a git pro

Issue #53, opened 2016-01-26
Resolving problems stemming from xray -> xarray transition


spencerahill:
A few problems for aospy have come up relating to the xray to xarray transition.
- Can't rename a dimension to it's existing name: https://github.com/pydata/xarray/issues/724
- The new xarray data model causes our `aospy.utils.replace_coord` function to break: https://github.com/pydata/xarray/issues/725

Both of them have workarounds that I've implemented in 38d2bcb.  The workaround for the latter is especially untenable in the long run, because the metadata for phalf gets kept for a pfull array.  So I'll leave this open to remind us to resolve them once they get fixed on the xarray side (or if we keep up with something smarter in the meantime).

spencerkclark:
:+1: I finally got around to updating to the latest release of xarray today; many thanks for these edits, and reaching out on the xarray side of things.

spencerahill:
No problem...I stumbled across these while trying to implement some upwind advection schemes.  In so doing, I decided to cleanup infinite-diff by using the xarray bug fixes, which led me down the rabbit hole :)

FYI I've now switched the `replace_coord` workaround with Stephan's suggestion of `to_dataset`.

spencerahill:
https://github.com/pydata/xarray/pull/736 solves the rename problem, so we can implement that after the next xarray release.

spencerahill:
https://github.com/pydata/xarray/pull/750 fixes the replece_coord problem, and has been included in xarray's 0.7.1 release, so we can now implement both these.  I will do so some time soon-ish.

spencerahill:
Both fixed in xarray 0.7.2 and implemented in #74.

Issue #32, opened 2015-11-06
Abstractify methods for finding data files saved on disk


spencerahill:
(Below is copied from @spencerkclark 's comment on #31 )

> While currently within Calc the two methods accomplish the same tasks, in an abstract sense the current ..._one_dir and ..._gfdl read-in methods are actually quite different:
> 
>    ..._one_dir requires the user to map every variable to a file name explicitly. If the variable does not appear in this map, aospy will not even attempt to look for it.
>    ..._gfdl is an implicit system. The mapping is coded into the method which looks for the files within the post-processing file structure. The user is not required a priori to specify which variables are in which files, and thus aospy is allowed to look for variables that may not exist.
> 
> I would argue that the explicit read-in method is the most general way of doing things. With enough information, one could automatically generate an explicit file map from an implicit generator. In addition, there is nothing that says you couldn't relax the current single directory constraint and just map each variable (within a particular time frequency) to a full file path.
> 
> To continue to support implicit read-in methods (for very structured output data, like ..._gfdl) you could require that a user create some object that implements an interface (call the interface FileMapGenerator?) to include methods to generate a map to files for a particular variable when given the intvl_in, variable name, data_in_dur etc. as arguments. The source code for these objects could be stored in a user's aospy_user directory.
> 
> Within a Run object one could then have a single argument for the file read-in method. The user could pass either the explicit dictionary mapping or they could pass an object that implements the FileMapGenerator interface. Within Calc, when reading in the files, you could have some simple logic that would be along the lines of: "if an explicit map is provided use the map; if not, use information about which variable you are looking for, and the interval in etc. and pass those as arguments to the generator, which would return an explicit map for just that variable." Using an interface would ensure that the explicit file map generated would always have the same structure (so that it could be used seamlessly within Calc).

spencerahill:
I really like this.  I agree, the implicit mapping should be specified by the user, since it no doubt varies so much among people (even within the lab; cf. #28 ).

[...]

PR #74, opened 2016-05-21
Misc. updates I've made over the past two months

spencerahill:
This is very poor form on my part to dump all these at once...got caught up in my thesis.  Taking a mini breather this afternoon and thought I'd push these.  To make matters worse, I don't have any tests.

@spencerkclark Would you mind checking out this branch (in both the informal and git senses)?  I'll update as needed to jive with #73 once it's merged (if the tests ever finish running ;)).  But just let me know if you see anything glaring.  Please wait until AFTER your generals exam...absolutely no rush on this

Re: backtracking on my previous pledge to not merge things in without tests: realistically it is not feasible for me to take the needed time to write tests until after August 2.  So I'll continue (with apologies) to be sloppy until then.  After the defense, in addition to the code revamp, I'd like to revamp the whole approach to this package, treating it like the legit open source, multi-user package it is, rather than my personal research tool that happens to be on github.

Some of the more relevant changes:
- Use `np.allclose` to determine if coordinates (namely `lat`) differ by very little between a Model and the given data; if they do then just overwrite the coord in the data with the Model's value
- Several bugfixes and added functionality in `plotting`
- Improve/make more consistent handling of logging/warnings
- Save region-specific pressure data for data on model coordinates
- Compute dt from the time bounds, rather than just assuming equal time spacing if the data lacks a dt array
- Bugfixes in `utils` `level_thickness` and `dp_from_ps` functions
- Bugfixes in surface area

### DIFF
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -10,16 +10,17 @@ from time import ctime
 import numpy as np
 import xarray as xr
 
-from . import Constant, Var
 from .__config__ import (LAT_STR, LON_STR, LAT_BOUNDS_STR, LON_BOUNDS_STR,
                          PHALF_STR, PFULL_STR, PLEVEL_STR, TIME_STR, YEAR_STR,
                          ETA_STR, BOUNDS_STR)
+from .constants import Constant, grav
 from .io import (_data_in_label, _data_out_label, _ens_label, _yr_label, dmget,
                  data_in_name_gfdl)
 from .timedate import TimeManager, _get_time
 from .utils import (get_parent_attr, apply_time_offset, monthly_mean_ts,
                     monthly_mean_at_each_ind, pfull_from_ps,
                     to_pfull_from_phalf, dp_from_ps, dp_from_p, int_dp_g)
+from .var import Var
 
 
 logging.basicConfig(level=logging.INFO)
@@ -672,11 +673,14 @@ class Calc(object):
         if zonal_asym:
             full_ts = full_ts - full_ts.mean(LON_STR)
         # Vertically integrate.
-        if self.dtype_out_vert == 'vert_int' and self.var.def_vert:
+        vert_types = ('vert_int', 'vert_av')
+        if self.dtype_out_vert in vert_types and self.var.def_vert:
             # Here we need file read-in dates (NOT xarray dates)
             full_ts = self._vert_int(full_ts, self._get_pressure_vals(
                 dp, self.start_date, self.end_date
             ))
+            if self.dtype_out_vert == 'vert_av':
+                full_ts *= (grav.value / self._to_desired_dates(self._ps_data))
         return full_ts, dt
 
     def _avg_by_year(self, arr, dt):
@@ -738,7 +742,7 @@ class Calc(object):
                         **{reg.name + '_pressure': coord}
                     )
             reg_dat.update(**{reg.name: data_out})
-        return reg_dat
+        return OrderedDict(sorted(reg_dat.items(), key=lambda t: t[0]))
 
     def _apply_all_time_reductions(self, full_ts, monthly_ts, eddy_ts):
         """Apply all requested time reductions to the data."""
@@ -759,7 +763,7 @@ class Calc(object):
                 reduced.update({reduc: self.region_calcs(data, func)})
             else:
                 reduced.update({reduc: self._time_reduce(data, func)})
-        return reduced
+        return OrderedDict(sorted(reduced.items(), key=lambda t: t[0]))
 
     def _make_full_mean_eddy_ts(self, data_in):
         """Create full, monthly-mean, and eddy timeseries of data."""
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -278,8 +278,12 @@ class Calc(object):
             dtype_lbl = dtype
         else:
             dtype = self.dtype_in_time
-        direc = os.path.join(data_in_direc, domain, dtype_lbl, self.intvl_in,
-                             str(self.data_in_dur[n]) + 'yr')
+        dur_str = str(self.data_in_dur[n]) + 'yr'
+        if self.dtype_in_time == 'av':
+            subdir = self.intvl_in + '_' + dur_str
+        else:
+            subdir = os.path.join(self.intvl_in, dur_str)
+        direc = os.path.join(data_in_direc, domain, dtype_lbl, subdir)
         files = [os.path.join(direc, data_in_name_gfdl(
                  name, domain, dtype, self.intvl_in, year, self.intvl_out,
                  self.data_in_start_date[n].year, self.data_in_dur[n]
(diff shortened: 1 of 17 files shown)


spencerkclark:
No worries -- I'll have a closer look a these changes sometime in the next few weeks.

spencerahill:
My remarkable git ineptitude persists (apologies), but the latest commits should close #53

aospy/calc.py
```diff
@@ -374,9 +375,11 @@ def _add_grid_attributes(self, ds, n=0):
                 except ValueError:
                     ds = ds
                 ds = ds.set_coords(name_int)
-                if not ds[name_int].equals(model_attr):
-                    warnings.warn("Model coordinates for '{}' "
-                                  "do not match those in Run".format(name_int))
+                if not np.array_equal(ds[name_int], model_attr):
```
spencerkclark, line 386: What was the reason for changing this from the DataArray equals method to the numpy array one?  The potential for slightly different coordinate indexes, but the same array values?  (I'm assuming you had a good reason for changing this; I'd just like to know for future reference).

aospy/calc.py
```diff
@@ -397,21 +400,11 @@ def _create_input_data_obj(self, var, start_date=False,
...
-        # calling xarray.concat?  Or does the year<1678 logic make this not
-        # possible?
-
-        # 2015-10-16 19:06:00 S. Clark: The year<1678 logic is independent of
-        # using xarray.open_mfdataset. The main reason I held off on using it
-        # here was that it opens a can of worms with regard to performance;
-        # we'd need to add some logic to make sure the data were chunked in a
-        # reasonable way (and logic to change the chunking if need be).
         for file_ in paths:
             test = xr.open_dataset(file_, decode_cf=False,
```
spencerkclark, line 459: Unrelated, but during the refactor we definitely need to rename this variable within the code (this was my doing in the very beginning...)

aospy/calc.py
```diff
@@ -787,7 +810,17 @@ def _load_from_scratch(self, dtype_out_time, dtype_out_vert=False,
...
-            return ds[region.name]
+            arr = ds[region.name]
+            # Use region-specific pressure values if available.
+            if self.dtype_in_vert == ETA_STR and not dtype_out_vert:
+                # try:
+                    arr[PFULL_STR].values = arr[region.name +
+                                                '_pressure'].values
+                # except KeyError:
+                    # return ds[region.name]
+                # else:
```
spencerkclark: Is there a reason to keep the commented-out code here?

aospy/calc.py
```diff
@@ -808,14 +810,15 @@ def _load_from_scratch(self, dtype_out_time, dtype_out_vert=False,
...
-                                     if r != PFULL_STR])
+                try:
+                    reg_pfull_str = region.name + '_pressure'
+                    arr = arr.drop([r for r in arr.coords.iterkeys()
+                                    if r not in (PFULL_STR, reg_pfull_str)])
+                except ValueError:
+                    return arr
+                else:
+                    arr = arr.rename({PFULL_STR: PFULL_STR + '_ref'})
+                    return arr.rename({reg_pfull_str: PFULL_STR})
```
spencerkclark, line 893: Sorry ignore my previous comment here...

aospy/calc.py
spencerkclark, line 386: Sorry, I really shouldn't be doing this commit by commit... I see what's happening here now.

spencerahill, line 386: Sorry yes no doubt I was not disciplined enough in how I organized my commits, so probably just looking at the final diff is the way to go.  Thanks a lot!

spencerahill, line 386: Take your time and I'll return to your comments after June 21.

aospy/model.py
```diff
@@ -72,14 +86,43 @@ def find_data_in_direc_repo(self, run_name='amip', var_name='ta',
 
         return glob.glob(os.path.join(direc_full, var_name + '_*.nc'))
 
+    def _get_grid_files_repo(self):
```
spencerkclark, line 89: I think we should eventually abstract this in a similar way that we are planning on abstracting the variable retrieval process (#32).  Right now it's a little too GFDL / CMIP5 specific (in terms of hard-coded directory and variable names) to reside in the aospy core.  This is fine for now though.

aospy/region.py
```diff
@@ -42,20 +44,35 @@ def mask_var(self, data):
 
     @staticmethod
     def _get_land_mask(data, do_land_mask):
+        if not do_land_mask:
+            return 1
+        try:
+            land_mask = data.land_mask.copy()
+        except AttributeError:
+            # TODO: Implement aospy built-in land mask to default to.
```
spencerkclark, line 52: How do you envision this being done?  Would we take a reference land mask and just regrid it to whatever resolution was needed by the particular run (based on its lat and lon coordinates)?

aospy/plotting.py
```diff
@@ -346,22 +345,39 @@ def _set_axes_labels(self):
             self.ax.set_title(self.ax_title, fontsize='small')
         # Axis panel labels, i.e. (a), (b), (c), etc.
         if self.ax_label:
+            if self.ax_label == 'auto':
+                text = '({})'.format(tuple('abcdefghijklmnopqrs')[self.ax_num])
```
spencerkclark: Could this line be simplified to something like this?

``` python
text = '({})'.format(string.ascii_letters[self.ax_num])
```

Albeit it would require an extra import statement for `string`; regardless I don't see the need to cast the string as a tuple.  String indexing should be fine here.

aospy/calc.py
```diff
@@ -432,68 +485,71 @@ def _create_input_data_obj(self, var, start_date=False,
...
-            raise ValueError("`dtype_in_vert` must be either 'pressure' or "
-                             "'sigma' for pressure data")
-        return data
+            return self._get_pressure_from_p_coords(ps, name=var.name, n=n)
+        if self.dtype_in_vert == ETA_STR:
+            return self._get_pressure_from_eta_coords(ps, name=var.name, n=n)
+        raise ValueError("`dtype_in_vert` must be either 'pressure' or "
+                         "'sigma' for pressure data")
 
     def _correct_gfdl_inst_time(self, arr):
```
spencerkclark, line 541: I think we'll eventually need a more transparent way of doing this (rather than do it behind the scenes); I'm assuming other modeling centers don't have this problem.  Currently it appears the only way to circumvent it for instantaneous data is to provide a time input interval string that does not end in 'hr'; is that really what we want?  No need to address this within this PR, but something to keep in mind for the refactor.

spencerahill:
@spencerkclark anything more to add? Otherwise I'll take a look at your comments over the coming days.  Thanks!

spencerkclark:
I'm all set.  Have a look whenever you are ready.

aospy/plotting.py
spencerahill: Nice, that's a great solution...wasn't aware of `string.ascii_letters`

aospy/region.py
spencerahill, line 52: Yes, that's pretty much what I had in mind.  I'll open a separate issue for this.

aospy/calc.py
spencerahill, line 541: Good call.  Ultimately we need to think about how to package together all of the site-specific stuff (e.g. for GFDL, this; dmget/hsmget).

We could potentially explicitly check the time indices compared to the nominal range given by the input file(s), but this could be tricky to implement.

aospy/model.py
spencerahill, line 89: Agreed.

spencerahill:
Thanks @spencerkclark!

Issue #33, opened 2015-11-06
Refactor: break Calc class up into smaller classes


spencerahill:
I have been thinking casually about this for a long time but was spurred by #32 to formalize it.  Even then, its still a work in progress.

Ultimately all of the filesystem IO should be factored out of `Calc` and into its own specialized classes.  Here's how I envision the overall flow, from user input to exporting the results:
- User desired input is parsed by some combination of classes (beyond the scope of this issue; see #4) to generate a `CalcSuite` #9 ).
- `CalcSuite` generates `CalcDataGetter` object, one per `Calc`. 
- Each `CalcDataGetter` generates `DataLoader` objects, one for each needed piece of data saved on disk.
- The `DataLoader` objects load their data from disk and creates `CalcData` objects.  To start, `CalcData` can basically just be a wrapper around `xray.DataArray`, with `xray_data` and `numpy_data` as attributes, holding the `xray.DataArray` and `np.ndarray` representations of the data, respectively.
- `CalcInterface` contains only the `xray.DataArray`s corresponding to the needed data and the minimal amount of additional metadata needed to execute the function.
- Time/space reductions are handled by a separate class, `Reduction`, that accepts a gridded timeseries and an object (its precise form TBD) that specifies all of the reductions to be performed. 
- `Reduction` generates a child class, `Reducer`, one per reduction to be performed, to execute them.  Some logic is necessary, or perhaps child classes `RegReducer` and `GriddedReducer`, 
- Each `Reducer`generates a `AOSData` object comprising its xray.DataArray and any other necessary metadata.
- Not sure which class should be responsible for this, but one of them then generates `DataExporter` objects that are responsible for writing the `AOSData` object to disk/database/whatever.

spencerahill:
Copying over relevant discussion w/ @spencerkclark in #90, wherein we have started in on a `DataLoader` module that handles loading data from disk.  Importantly, `DataLoader` will be glued on at the `Run` level rather than to each `Calc`: each `Run` will have a `DataLoader` object as an attribute, and whenever data is to be loaded from disk corresponding to that `Run`, the `DataLoader` will be used.  This is a departure from the `CalcDataGetter` and `DataLoader` steps described above, although we may still end up using the former.

> > Will we need to implement the GFDL and one_dir mappings before we can meaningfully attempt plugging this into Calc?
> 
[...]

PR #90, opened 2016-09-27
WIP: Refactor variable-loading portion of calc.py

spencerkclark:
This PR is a start on addressing a number of issues (most notably #32 and #33).  This is very much a work in progress, but I wanted to get some discussion started on this topic.  @spencerahill when you get a chance, please have a look!

At the core of this PR is a new `DataLoader` base class, which implements some methods (with unit tests) meant to streamline and improve upon the current setup in in [Calc._create_input_data_obj](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L442) and [Calc._add_grid_attributes](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L376).   The `DataLoader.load_variable` method is what I envision would be called in `calc.py` directly (and replace a lot of existing code).  It takes the requested variable, start date, end date, and interval in as arguments, and returns a DataArray with the specified variable, subset for the requested time range, with all local grid attributes renamed to match `aospy` internal conventions.

I'll emphasize here that currently, both `Calc._add_grid_attributes` and `Calc._create_input_data_obj`  do a lot:
- `Calc._add_grid_attributes` not only renames local grid attributes, but also assimilates `Model` grid attributes into the existing Dataset. 
- `Calc._create_input_data_obj` also computes a `dt` and `pfull` variable if required and corrects out of range dates if needed. 

I currently do not implement any of those things in `DataLoader`.  The only one I think that would make sense to include eventually would be the date range adjustment step (it would be part of the `preprocess` method called by `xr.open_mfdataset` in `DataLoader._sel_var`).  Assimilating `Model` grid attributes and computing pressures and timedeltas can be done elsewhere if needed.

Methods called by `DataLoader.load_variable` are meant to improve upon the current setup.  To start, there is no need to manually concatenate the files we read in; in fact we should leverage `xr.open_mfdataset`'s lazy-loading functionality to prevent unnecessarily loading in datapoints from un-needed date ranges and un-needed variables.  This is done in the `DataLoader._sel_var` method.  It loads a set of files, preprocesses them to rename the grid attributes, and then searches the Dataset for the requested variable, without loading anything into memory.  The `DataLoader._sel_time` method then first checks to see if data for the requested variable exist for the full time range requested (addressing #72), before returning the sliced DataArray.  Finally, before returning, the `DataLoader.load_variable` method explicitly "loads" the data into memory (i.e. converts the underlying array from a dask array to a numpy array) to prevent previously-encountered chunking-related performance issues.

Here, an extremely basic `DataLoader` is implemented for testing purposes.  It is called a `DictDataLoader`, because it simply takes a `dict` as the sole argument in its constructor, which maps a string `intvl_in` to a list of filepaths.  It is assumed that all variables the user may request can be found in those files.  An example can be found in the unit tests of the PR. 

I have not yet ported the current methods (e.g. `'gfdl'`, `'one_dir'` etc.) to this paradigm yet.  That being said, the way things are designed here, it would just require some new classes, extending the base `DataLoader` class, each implementing their own `DataLoader._generate_file_set` method (which generates a list of filepaths containing data for a given variable, time range, and input interval type) consistent with how they currently operate.

PS: Apologies for the odd commit history here -- the diff reflects only the changes associated with this PR (i.e. the commit on Sept. 27).  If we eventually elect to merge a later version of this, I'll try and clean things up.

Remaining Work Needed
-------------------------
- [x] Merge the result of #97 (in particular the changes to what is currently the ```timedate.py``` module)
- [x] Convert the general custom ```correction``` method prior to time-subsetting to a specific "time-shift" correction that will enable the option of something along the lines of what is currently ```Calc._correct_gfdl_inst_time```
- [x] Add something akin to the "copy-and-override" method
- [x] Implement the ```GFDLDataLoader```
- [x] Implement the functionality of a ```OneDirDataLoader```
- [x] Integrate DataLoaders into calc.py
- [x] Temporarily remove code that will need to be repurposed later as a ```GFDLRepoDataLoader```
- [x] Add documentation and examples for the DataLoaders
- [x] Add comprehensive tests of all added methods

Implementing a DataLoader that will allow one to select a file set based on a number of different parameters (not just with a single key identifier) will probably need to wait until later.


spencerahill:
Very excited about this!  Tied up today, but should have time before the weekend to give this a full review.

spencerahill:
I am getting errors when I try to run the test suite that ultimately stem from the import statements within `test/test_objs`.  Specifically, I have to change the local imports to explicit relative imports, i.e. `import runs` in `test/test_objs/models.py` has to be changed to `from . import runs` to work on my Macbook.  Does that happen for you?

Those lines aren't part of your PR, just wanted to make note of it.  I'll push commits with those fixed.  (Example of something that would have been caught already if CI was configured properly).

Now on to the actual PR :)

aospy/data_loader.py
```diff
@@ -0,0 +1,168 @@
+"""asopy DataLoader objects"""
```
spencerahill: typo

aospy/data_loader.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class DataLoader(object):
+    """A fundamental DataLoader object"""
+    def __init__(self):
+        self.correction = None
+
+    def load_variable(self, var=None, start_date=None, end_date=None,
+                      intvl_in=None):
+        """Returns the DataArray with requested variable, given time range,
```
spencerahill: Style nit: here and in other docstrings, make the verbs imperative, i.e. "Return the DataArray" rather than "Returns".  (It's possible I have not been consistent on this myself in other docstrings, but let's do this moving forward.)

spencerkclark:
Yeah, we should probably sort that out -- I can't seem to reproduce that behavior though.  How are you running the tests?

spencerahill:
@spencerkclark Just with py.test from the command line, i.e. `py.test /path/to/aospy/test`.

Maybe it stems from me using PYTHONPATH for aospy, rather than formally installing it via setup.py?

spencerkclark:
Odd, I'm running them the same way with aospy in my PYTHONPATH as well (i.e. not formally installed).

aospy/data_loader.py
```diff
@@ -0,0 +1,168 @@
...
+    def _generate_file_set(self, var, start_date, end_date, intvl_in):
+        raise NotImplementedError(
+            'All DataLoaders require a _generate_file_set method')
+
+
+class DictDataLoader(DataLoader):
+    """A data loader that corresponds with a dictionary mapping lists of files
+    to interval in tags.
+    """
+    def __init__(self, file_map={}, correction=None):
```
spencerahill: I am definitely guilty of doing this in the past, but we should avoid mutables as default values moving forward: http://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

spencerahill:
I still need to savor the internals, so this is mostly nits for now.  But I certainly don't see any glaring problems.  When we chat Wednesday I should have a better handle of it all and thus more substantive comments.

aospy/__config__.py
```diff
@@ -16,5 +17,41 @@
...
+AVERAGE_DT_STR = 'average_DT'
+NV_STR = 'nv'
+AVG_START_DATE_STR = 'avg_start_date'
+AVG_END_DATE_STR = 'avg_end_date'
+
+# AVERAGE_DT_STR does not enter here, because it is not a 'time since'
+# quantity.
+TIME_VAR_STRS = [TIME_STR, TIME_BOUNDS_STR]
+
+GRID_ATTRS = OrderedDict(
```
spencerahill: Thank you...I've been meaning to do something like this for a while!

aospy/__init__.py
```diff
@@ -1,6 +1,9 @@
 """aospy: management, analysis, and plotting of gridded climate data."""
 from .__config__ import (user_path, LAT_STR, LON_STR, PFULL_STR, PHALF_STR,
```
spencerahill: I think we should consider having a module or dictionary with all the _STR variables.  That way the import could be a single object.  I.e. if it was a module called `internal_names` then we could do `from . import internal_names` and then `internal_names.LAT_STR` etc.  That would clean up import statements in a lot of places.

aospy/data_loader.py
```diff
@@ -0,0 +1,168 @@
...
+        file_map : dict
+            A dict mapping an input interval to a list of files
+        correction : function
+            An optional function to correct an aspect of input data before
+            subsetting in time
+        """
+        self.file_map = file_map
+        self.correction = correction
+
+    def _generate_file_set(self, var, start_date, end_date, intvl_in):
```
spencerahill: Seems a bit clunky to have start_date and end_date as required but unused args that go before intvl_in.  Is there potentially a better call signature?  Tricky because the requirements/logic will vary across DataLoader implementations

spencerahill:
Will we need to implement the GFDL and one_dir mappings before we can meaningfully attempt plugging this into Calc?

spencerkclark:
@spencerahill Thanks for taking a first look at this.  I think I should be able to address most of your initial review comments (all very good suggestions) before we talk on Wednesday.

> Will we need to implement the GFDL and one_dir mappings before we can meaningfully attempt plugging this into Calc?

The short answer is yes.  Unfortunately it may take a fair amount more, since as it stands now, Calc is quite convoluted (and not really atomic enough to simply plug this logic in on its own).  I'll start a list of a few things here, but there are probably still other things we'll need to consider:
- `_add_grid_attributes` in Calc currently does what `rename_grid_attrs` and `set_grid_attrs_as_coords` do in DataLoader and more.  It also compares grid attributes to those of the associated Model (how do we want to maintain this behavior moving forward?, see discussion in #14). 
- The effects of `dt_from_time_bounds` and `_get_dt` are both now implemented in the timedate module.  I don't think we should require Calc to have a `dt` attribute anymore.  It should just be carried with the DataArrays where needed as a coordinate.  If the DataArray does not have an `'average_DT'` attribute, then one can just take a time mean without weighting by `dt` (thus there is no reason to generate an array of ones for instantaneous data, [as is done now](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L645)).
- In `DataLoader.load_variable` I subset things immediately in time to isolate the full time series of the data of interest (before loading it into memory).  This and further time subsetting of data (i.e. the selection of a series of months for seasonal averages) is currently handled in the `_to_desired_dates` function.
- Tracing all the way back to a main-like script, what is the minimum set of parameters needed to identify a given file set for any DataLoader?  How should we specify those parameters when submitting a computation?  In some ways this traces back to your comment on `_generate_file_set` above.  I have some ideas, but it's probably best we discuss those verbally on Wednesday.

Feel free to list any other concerns of your own that come to mind!

aospy/data_loader.py
```diff
@@ -0,0 +1,180 @@
...
+                               decode_cf=False)
+        ds = TimeManager.enforce_valid_timestamp_date_range(ds)  # Tested
+        if internal_names.TIME_BOUNDS_STR in ds:
+            ds = TimeManager.set_average_dt_metadata(ds)  # Tested
+        ds = self.set_grid_attrs_as_coords(ds)  # Tested
+        ds = xr.decode_cf(ds)
+        da = self._sel_var(ds, var)  # Tested
+
+        # Apply correction before selecting time range.
+        if self.correction:
```
spencerahill: Let's rename this to something more descriptive and general.

aospy/data_loader.py
```diff
@@ -0,0 +1,180 @@
...
+             interval coding (e.g. 'monthly')
+
+        Returns
+        -------
+        da : DataArray
+             DataArray for the specified variable, date range, and interval in
+        """
+        file_set = self._generate_file_set(var, start_date, end_date, intvl_in,
+                                           vert_in)
+        ds = xr.open_mfdataset(file_set, preprocess=self.rename_grid_attrs,
```
spencerahill: Let's make a method for this step within DataLoader, i.e. `def _get_data_from_disk` (maybe something else?) for the sake of modularity.

aospy/data_loader.py
```diff
@@ -0,0 +1,180 @@
...
+                               concat_dim=internal_names.TIME_STR,
+                               decode_cf=False)
+        ds = TimeManager.enforce_valid_timestamp_date_range(ds)  # Tested
```
spencerahill: Put these three lines into their own internal method, and potentially add the `xr.decode_cf` step here if (1) it doesn't mess things up to put that before the `self.set_grid_attrs_as_coords` step and (2) `decode_cf` doesn't modify anything besides the time-related quantities.

spencerkclark:
Some relevant discussion took place in #97 and offline regarding the specification of ```DataLoader```'s in the constructors of ```Proj```, ```Model```, and ```Run``` objects; I'll do my best to summarize things here.  @spencerahill feel free to chime in if I'm missing anything.

Usage of DataLoaders with aospy core objects
-----------------------------------------------
The way aospy currently works is that metadata (e.g. the ```data_*``` arguments) for all possible data loading methods (e.g. ```'one_dir'```, ```'gfdl'```) can be specified as keyword arguments directly in the constructors of all aospy core objects (```Proj```, ```Model```, or ```Run```).  This is one way to allow for the "bottom-up" specification of attributes (i.e. if an attribute is not specified in a Run's constructor, the Run's Model is checked for that attribute, and so on).  This approach (specifying DataLoader-specific attributes as arguments directly in aospy core objects) is not ideal for two reasons:
1. A user could specify an attribute in a constructor that will do nothing (e.g. if one specifies a ```data_start_date```, but uses the ```'one_dir'``` method)
2. From a documentation perspective, if we intend to support an arbitrary number of DataLoaders, with wide ranging needed arguments to their constructors, we would need to specify a long list of possible keyword arguments to all aospy core objects, with long documentation about which ones corresponded to which data loading method.  

This suggests the need for an alternative.  With this PR, it will be possible for aospy core objects to take an individual DataLoader object as a single keyword argument.  Arguments needed for the construction of the specific DataLoader used would be provided directly in the DataLoader constructor (rather than in the core objects).  This would solve both of the two issues above -- it would no longer be possible to specify powerless keyword arguments, and all documentation of DataLoader-specific attributes could be handled at the level of the individual DataLoaders.  

Inheritance of DataLoader attributes would not work quite like before, but if there was significant overlap between DataLoader parameters for runs of a single Model, one could specify a DataLoader of the needed type in the Model object, and override the remaining parameters on a Run-by-Run basis.  For example:
```python
am2 = Model(
    name='am2',
    data_loader=GFDLDataLoader(
        data_start_date=datetime.datetime(2000, 1, 1),
        data_end_date=datetime.datetime(2099, 12, 31),
        data_in_dir='path/to/data',
    )
)

control = Run(
    name='control',
    data_loader=am2.data_loader.override(data_in_dir='/path/to/data/control')
)
```
In this case the control Run would inherit the ```data_start_date``` and ```data_end_date``` attributes from the ```am2``` Model's DataLoader, but modify the ```data_in_dir``` attribute.  

This is sort of the track that this PR has been going down already, but I think it is important to try and flesh out how we both think these objects will be used in the future.

spencerahill:
@spencerkclark just a ping that I have merged #90, so that's no longer a blocker.  I'm done for the night but I'll get back to you on your most recent comment.  Thanks!

spencerahill:
@spencerkclark the summary looks good to me, thanks!

>I think it is important to try and flesh out how we both think these objects will be used in the future.

Honestly I can't think of any major problems that would arise here.  This will make things more explicit, modular, and easily testable moving forward.  The only downside is the slightly more involved call-signature of the Proj/Model/Run __init__ methods, but I think it's worth it.  So IMO full speed ahead!

The only thing I'd say is, given our intentions with #111, let's make sure that there's no logic that assumes anything about the type of an object's parent (not that there is any yet, haven't looked).

spencerahill:
This will also help reduce boilerplate.  For example, if for a single Model you have, in addition to Runs ('a', b', 'c', ...) that use that model's Data Loader, another set of runs ('aa', 'bb', 'cc', ...) that use a different set of parameter (e.g. different date range), you can specify the `DataLoader.override` in only one of the Runs ('aa') and then have the rest ('bb', 'cc', ...) just reference that Run's DataLoader.

This is a real use-case for me; my AM2 model object has several large sets of runs, each with different date ranges.

aospy/utils/times.py
```diff
@@ -290,3 +291,150 @@ def extract_date_range_and_months(time, start_date, end_date, months):
...
+            except pd.tslib.OutOfBoundsDatetime:
+                year = pd.Timestamp.min.year + 2
+                # TODO: don't hard code Jan. 1 as the reference date.
+                units_str = '{0} since {1}-01-01 00:00:00'.format(freq,
+                                                                  year)
+                ds[VAR_STR].attrs['units'] = units_str
+    return ds
+
+
+def set_average_dt_metadata(ds):
```
spencerkclark: In xarray we are considering a ```NetCDFTimeIndex```, which would already simplify a lot of the logic in this PR -- notably not this function, however.  Maybe I'm getting ahead of myself, but perhaps there could also be interest in a ```NetCDFTimePeriodIndex```?  If managed correctly, I think this could be used to more idiomatically (in terms of [pandas](http://pandas.pydata.org/pandas-docs/stable/timeseries.html#periodindex-and-period-range) / xarray parlance) solve the problem (#72) that required this function to solve.

E.g. with existing pandas code one can do:
```python
In [1]: import pandas as pd

In [2]: rng = pd.period_range('2000-01-01', periods=3, freq='M')

In [3]: rng
Out[3]: PeriodIndex(['2000-01', '2000-02', '2000-03'], dtype='int64', freq='M')

In [4]: rng[0].start_time
Out[4]: Timestamp('2000-01-01 00:00:00')

In [5]: rng[-1].end_time
Out[5]: Timestamp('2000-03-31 23:59:59.999999999')
```
Therefore instead of having to track the absolute time bounds of a dataset ourselves, we could let a ```PeriodIndex``` do things in the background and pick off the start and end times as needed.  Obviously instead of ```Timestamp``` objects being returned, in the case of unusual date ranges / calendars, we'd want ```netcdftime``` datetime objects.

spencerahill:
FYI don't worry about the failing appveyor tests on this branch

spencerkclark:
@spencerahill I feel like I'm making progress here, though right now my lack of familiarity with the ```'gfdl_repo'``` method of loading files is holding me back.  I'm having a hard time parsing which parameters are set permanently and which are set at runtime (i.e. when one submits a calculation).  This distinction dictates which parameters are set in the DataLoader constructor, and which must get passed dynamically to ```DataLoader.load_variable``` and ultimately ```DataLoader._generate_file_set```.  

Do you think we can punt on that for now -- meaning temporarily take it out completely and reimplement it later?

Edit:  I think what I suggest above is a little extreme.  I think what's mainly throwing me off is actually the fact that grid files are loaded dynamically using the GFDL repo method (this is different than the explicit specification of grid files in all other methods).  That said, I think we could probably work something out.  I'd just might need your help to talk it through a bit.

spencerahill:
The way I coded it is likely opaque, but the directory structure itself is straightforward.  See e.g. `/archive/Spencer.Hill/am2/am2clim_reyoi/gfdl.ncrc3-intel-prod/pp/` for an example.  The model, run, dtype_in (e.g. `av` or `ts`), variable name, time interval in (e.g. 3hr or monthly), date range, and duration of each netCDF file (e.g. `../ts/monthly/30yr` or `../av/annual_15yr`) are all required to find the needed netCDF files.

Of these, it would seem to me that only the run, model, and date range (spanned by all the data, not necessarily the date range the user wants to compute over) would make sense to specify beforehand for each DataLoader...I can think of use-cases where it would be desirable to specify each of the other parameters at runtime.

I rely on `gfdl_repo` for most of my simulations, so I would prefer not to do away with it.  But I also don't want it to be a bottleneck for the rest of this work.  

What do you think?  I'd be willing to try to code up a DataLoader for it.  Are you thinking we should get 100% rid of the legacy code for data loading in this PR (in which case the GFDL DataLoader is a bottleneck)?

spencerkclark:
Sorry if I was confusing -- I meant the [```'gfdl_repo'``` functionality](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L279) (i.e. for loading CMIP data).  The [```GFDLDataLoader``` (for the directory structure you just described)](https://github.com/spencerahill/aospy/pull/90/files#diff-c1bc79d9c0310c1c1c15a5943463101cR272) is up and running already.  The path-construction part of the code could use some tests, but I basically just ported the old functions over.

I've started a new branch in my object library, with some real-life examples using a [```DictDataLoader```](https://github.com/spencerkclark/aospy-obj-lib/blob/aospy-beta/aospy_user/runs/imr_aospy_beta.py) and [```GFDLDataLoader```](https://github.com/spencerkclark/aospy-obj-lib/blob/aospy-beta/aospy_user/runs/cases.py) and both are working for real life data and variables.

spencerahill:
My mistake!  You were clear -- I just wasn't paying close enough attention.

Yes, we can remove the `gfdl_repo` for now.  I actually will have a need for it again within the next month or so, but I would like to create the DataLoader for it myself when that time comes...will help make sure I understand the codebase.

spencerkclark:
> Yes, we can remove the gfdl_repo for now. I actually will have a need for it again within the next month or so, but I would like to create the DataLoader for it myself when that time comes...will help make sure I understand the codebase

Sounds good, I just wanted to clarify the priorities for v0.1 (i.e. how much we should push for certain things versus others).  Like I mentioned in my edit last night, I think it's doable, but I would need a little guidance.  Pending tests of path generation for the GFDLDataLoader and temporarily pruning code for the ```'gfdl_repo'``` method, this PR is pretty much set for your review, so I would say it would be great if you could start taking a closer look at things when you get a chance (I realize there's a lot to look over).

spencerahill:
Excellent.  May not start in on the review today, but certainly will do it within the next couple days.  Thanks for all the hard work.

spencerkclark:
>Pending tests of path generation for the GFDLDataLoader and temporarily pruning code for the 'gfdl_repo' method

@spencerahill tests of path generation for the GFDLDataLoader have been added and the ```'gfdl_repo'``` method for loading files has been temporarily removed.  

For future reference (when looking to reimplement the functionality as a ```GFDLRepoDataLoader```) 2f96841 and 030f355 are the commits that remove the ```'gfdl_repo'``` method of loading files.

aospy/__config__.py
```diff
@@ -1,4 +1,5 @@
 import os
```
spencerahill, line 1: The internal names are now defined both here and in `internal_names`.  Is there a reason for this?  If not let's have them in internal_names only.

aospy/__init__.py
```diff
@@ -1,6 +1,9 @@
 """aospy: management, and analysis of gridded climate data."""
 from .__config__ import (user_path, LAT_STR, LON_STR, PFULL_STR, PHALF_STR,
```
spencerahill: Modify this pending above comment on internal_names v. __config__

spencerahill: Somewhat of a tangent, but do you think we need to import all the names to the package-level namespace?  It seems to me that leaving them in internal_names and only doing `from . import internal_names` would be cleaner.  (I know you weren't responsible for this in the first place)

aospy/calc.py
```diff
@@ -152,6 +145,14 @@ def __init__(self, proj=None, model=None, run=None, ens_mem=None, var=None,
         self.end_date_xarray = (self.start_date_xarray +
                                 (self.end_date - self.start_date))
 
+        # Add DataLoader logic
```
spencerahill: I would omit this comment; the code itself is self explanatory.

aospy/calc.py
```diff
@@ -244,132 +245,11 @@ def __init__(self, calc_interface):
...
-        else:
-            msg = ("netCDF files for calc object `{0}`, variable `{1}`, year "
-                   "range {2}-{3}, in directory {4}, not found")
-            raise IOError(msg.format(self, var, start_date, end_date,
-                                     data_direc))
-        paths = list(set(files))
-        paths.sort()
-        return paths
-
+    # 2017-01-13 [SKC]
```
spencerahill: Moving forward, let's get out of the habit (that I started) of leaving detailed inline comments of future TODOs.  Instead, let's create an Issue for each one.  If the inline stuff is necessary in order to understand the logic, then it can also be copied into the method's docstring (omitting the date and signature).

aospy/calc.py
```diff
@@ -562,17 +366,36 @@ def _get_input_data(self, var, start_date, end_date, n):
             set_dt = True if not hasattr(self, 'dt') else False
             cond_pfull = ((not hasattr(self, 'pfull')) and var.def_vert and
                           self.dtype_in_vert == ETA_STR)
-            data = self._create_input_data_obj(var, start_date, end_date, n=n,
-                                               set_dt=set_dt,
-                                               set_pfull=cond_pfull)
+            data = self.data_loader.load_variable(var, start_date, end_date,
+                                                  self.time_offset,
+                                                  **self.data_loader_attrs)
+            # 2017-01-13 [SKC]: Load variable returns a DataArray (for now)
```
spencerahill: Same as above comment

aospy/calc.py
```diff
@@ -14,6 +14,7 @@
 from .__config__ import (LAT_STR, LON_STR, LAT_BOUNDS_STR, LON_BOUNDS_STR,
```
spencerahill: Remove unused `import pandas as pd` and `BOUNDS_STR`

aospy/internal_names.py
```diff
@@ -0,0 +1,56 @@
+"""Internal names used throughout aospy"""
```
spencerahill: Can we flesh out this docstring w/ a few sentences after the 1-line summary.  Something along the lines of "Depending on its source, data corresponding to the same physical variable can have different internal names.  To ensure that all variables can be easily accessed, aospy overwrites these names with the names specified in this module."

spencerahill: And maybe explicitly refer to `GRID_ATTRS`, e.g. for each physical variable, aospy stores a list of potential names (based on all previously found names of that variable in different datasets).  aospy then finds that variable by searching for those variable names.

aospy/model.py
```diff
@@ -14,36 +15,33 @@
 from .utils.times import datetime_or_default
 from . import utils
 
+# Need to set dask to use the serial scheduler in order to use multiprocess
```
spencerahill: This wording is a little clunky...maybe "dask must use its serial scheduler if computations are to be performed in parallel using multiprocess" (here and elsewhere this occurs)

aospy/model.py
```diff
@@ -14,36 +15,33 @@
...
-        self.data_dir_struc = data_dir_struc
-        self.data_dur = data_dur
 
-        self.data_start_date = datetime_or_default(data_start_date, None)
-        self.data_end_date = datetime_or_default(data_end_date, None)
-        self.default_start_date = datetime_or_default(default_start_date,
-                                                      self.data_start_date)
-        self.default_end_date = datetime_or_default(default_end_date,
-                                                    self.data_end_date)
+        try:
```
spencerahill: Why do we need this try/except guard now?

aospy/run.py
```diff
@@ -2,49 +2,31 @@
...
-        self.data_suffix = data_suffix
-        self.data_files = {} if data_files is None else data_files
-
-        self.data_start_date = datetime_or_default(data_start_date, None)
-        self.data_end_date = datetime_or_default(data_end_date, None)
-        self.default_start_date = datetime_or_default(default_start_date,
-                                                      self.data_start_date)
-        self.default_end_date = datetime_or_default(default_end_date,
-                                                    self.data_end_date)
+        try:
```
spencerahill: Same question as above

aospy/test/data/objects/examples.py
```diff
@@ -1,23 +1,24 @@
 import os
```
spencerahill, line 1: Just out of curiosity, do you know why the example netCDF files are showing up as changed?

aospy/test/data/objects/examples.py
```diff
@@ -1,23 +1,24 @@
...
+from aospy.data_loader import DictDataLoader
 
 ROOT_PATH = os.path.dirname(__file__)
 
 
 def total_precipitation(convection_rain, condensation_rain):
     return convection_rain + condensation_rain
 
-files = ['{:04d}0101.precip_monthly.nc'.format(year) for year in range(4, 7)]
+files = os.path.join(os.path.join(os.path.split(ROOT_PATH)[0], 'netcdf'),
```
spencerahill: I think you can do `os.path.join(os.path.split(ROOT_PATH)[0], 'netcdf', '000[4-6]0101.precip_monthly.nc')`

aospy/test/test_data_loader.py
```diff
@@ -0,0 +1,392 @@
+#!/usr/bin/env python
+"""Test suite for aospy.data_loader module."""
+import unittest
+from datetime import datetime
+import xarray as xr
```
spencerahill: blank line between python internal and external packages.  Also let's alphabetize imports within each category

aospy/test/test_data_loader.py
```diff
@@ -0,0 +1,392 @@
...
+        assert self.ALT_LAT_STR in self.ds
+        ds = self.DataLoader.rename_grid_attrs(self.ds)
+        assert LAT_STR in ds
+
+    def test_set_grid_attrs_as_coords(self):
+        ds = self.DataLoader.rename_grid_attrs(self.ds)
+        sfc_area = ds[self.var_name].isel(**{TIME_STR: 0}).drop(TIME_STR)
+        ds[SFC_AREA_STR] = sfc_area
+
+        # Assert that SFC_AREA_STR is not initially carried by self.var_name
```
spencerahill: This comment and the one's like it aren't needed; the code is self-explanatory

aospy/test/test_data_loader.py
```diff
@@ -0,0 +1,392 @@
...
+        ds = xr.decode_cf(self.ds)
+        da = ds[self.var_name]
+
+        # If you don't pass a time_offset or non-null DataVars,
+        # nothing should happen
+        result = self.DataLoader._maybe_apply_time_shift(da.copy())[TIME_STR]
+        assert result.identical(da[TIME_STR])
+
+        # If you do pass an offset, it should be applied
+        # Why does da get modified in place?
```
spencerahill: Omit this question.  Are you asking why you have to do da.copy?

aospy/test/test_data_loader.py
```diff
@@ -0,0 +1,392 @@
...
+        self.assertEqual(result, expected)
+
+        expected = ['/test/atmos_inst/ts/monthly/'
+                    '6yr/atmos_inst.200601-201112.temp.nc']
+        result = self.DataLoader._input_data_paths_gfdl(
+            'temp', datetime(2010, 1, 1), datetime(2010, 12, 31), 'atmos',
+            'monthly', 'pressure', 'inst', None)
+        self.assertEqual(result, expected)
+
+        # 2017-01-16 SKC: this test fails; I'm not sure what the result
```
spencerahill: Not sure I follow this one...I've never seen 'monthly_from_daily' before.  Have you seen it before?  Otherwise not sure we need to test for it.

aospy/test/test_data_loader.py
```diff
@@ -0,0 +1,392 @@
...
+            io.data_name_gfdl('temp', 'atmos', 'av',
+                              '3hr', 2010, None, 2000, 6)
+
+        # Test 'av_ts' data
+        expected = 'atmos.2006-2011.01-12.nc'
+        result = io.data_name_gfdl('temp', 'atmos', 'av_ts',
+                                   '3hr', 2010, None, 2000, 6)
+        self.assertEqual(result, expected)
+
+    @unittest.expectedFailure  # This aspect of this function is obselete
```
spencerahill: Sorry, what do you mean here?

aospy/test/test_utils_io.py
```diff
@@ -23,5 +23,10 @@ def test_dmget(self):
                   'gfdl.ncrc3-default-repro/1/history/'
                   '00010101.atmos_month.nc'])
 
+        # Test with a pure string argument
+        io.dmget('/home/Spencer.Clark/archive/imr_skc/control/'
+                 'gfdl.ncrc3-default-repro/1/history/'
+                 '00010101.atmos_month.nc')
+
 if __name__ == '__main__':
```
spencerahill, line 28: Add one more blank line

aospy/utils/times.py
```diff
@@ -6,6 +6,7 @@
 import xarray as xr
 
 from ..__config__ import TIME_STR
```
spencerahill: Same as above re: config v. internal_names

aospy/utils/times.py
```diff
@@ -322,3 +328,115 @@ def extract_date_range_and_months(time, start_date, end_date, months):
     inds &= (time[TIME_STR] <= np.datetime64(end_date))
     inds &= (time[TIME_STR] >= np.datetime64(start_date))
     return time.sel(time=inds)
+
+
+def set_average_dt_metadata(ds):
+    """If the Dataset or DataArray contains time average data, enforce
```
spencerahill: Start docstring w/ one-line summary

aospy/utils/times.py
```diff
@@ -322,3 +328,115 @@ def extract_date_range_and_months(time, start_date, end_date, months):
...
+    that there are coordinates that track the lower and upper bounds of
+    the time intervals, and that there is a coordinate that tracks the
+    amount of time per time average interval.
+
+    CF conventions require that a quantity stored as time averages
```
spencerahill, line 342: Might be nice to provide link to official CF conventions on this point.

aospy/utils/times.py
```diff
@@ -322,3 +328,115 @@ def extract_date_range_and_months(time, start_date, end_date, months):
     inds &= (time[TIME_STR] <= np.datetime64(end_date))
     inds &= (time[TIME_STR] >= np.datetime64(start_date))
     return time.sel(time=inds)
+
+
+def set_average_dt_metadata(ds):
```
spencerahill: Maybe include cf explicitly in function name, e.g. `ensure_time_avg_has_cf_metadata` (or something else)...`set_av_data_cf_metadata`?

aospy/utils/times.py
```diff
@@ -322,3 +328,115 @@ def extract_date_range_and_months(time, start_date, end_date, months):
...
+    # For PEP8 line-length reasons, make these local variables.
+    AVG_START_DATE_STR = internal_names.AVG_START_DATE_STR
+    AVG_END_DATE_STR = internal_names.AVG_END_DATE_STR
+    TIME_BOUNDS_STR = internal_names.TIME_BOUNDS_STR
+    TIME_STR = internal_names.TIME_STR
+    NV_STR = internal_names.NV_STR
+    AVERAGE_DT_STR = internal_names.AVERAGE_DT_STR
+
+    if AVERAGE_DT_STR not in ds:
+        # Note average_DT will NOT have a units attribute and therefore
```
spencerahill: I haven't thought about this.  If you can see a plausible use case, move this to an Issue.  Otherwise remove.

aospy/utils/times.py
```diff
@@ -322,3 +328,115 @@ def extract_date_range_and_months(time, start_date, end_date, months):
...
+        times = da.time.isel(**{internal_names.TIME_STR: [0, -1]})
+        da_start, da_end = times.values
+    message = 'Data do not exist for requested time range: {0} to {1}'
+    range_exists = start_date >= da_start and end_date <= da_end
+    assert (range_exists), message.format(start_date, end_date)
+
+
+def sel_time(da, start_date, end_date):
+    """Subset a DataArray or Dataset for a given date range.  Ensures
+    that data are present for full extend of requested range.
```
spencerahill: typo 'extend'

aospy/utils/times.py
```diff
@@ -322,3 +328,115 @@ def extract_date_range_and_months(time, start_date, end_date, months):
...
+    AssertionError
+         if the time range is not within the time range of the DataArray
+    """
+    if internal_names.AVG_START_DATE_STR in da:
+        da_start = da[internal_names.AVG_START_DATE_STR].values
+        da_end = da[internal_names.AVG_END_DATE_STR].values
+    else:
+        times = da.time.isel(**{internal_names.TIME_STR: [0, -1]})
+        da_start, da_end = times.values
+    message = 'Data do not exist for requested time range: {0} to {1}'
```
spencerahill: 'does' reads better to me than 'do' (even if the latter is grammatically correct since data is technically plural here)

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+# Need to set dask to use the serial scheduler in order to use multiprocess
+# to compute calculations in parallel
+dask.set_options(get=dask.async.get_sync)
+
+
+class DataLoader(object):
+    """A fundamental DataLoader object"""
+    def load_variable(self, var=None, start_date=None, end_date=None,
+                      time_offset=None, **DataAttrs):
+        """Return a DataArray with requested variable, given time range,
```
spencerahill: Start w/ 1-line summary

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        and input interval.
+
+        This function automatically renames all grid attributes to match
```
spencerahill: Can omit 'this function'

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+             DataArray for the specified variable, date range, and interval in
+        """
+        file_set = self._generate_file_set(var=var, start_date=start_date,
+                                           end_date=end_date, **DataAttrs)
+        ds = self._load_data_from_disk(file_set)
+        ds = self._prep_time_data(ds)
+        ds = self.set_grid_attrs_as_coords(ds)  # Tested
+        da = self._sel_var(ds, var)  # Tested
+
+        # Apply correction before selecting time range.
```
spencerahill: move to Issue

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        # override _maybe_apply_time_shift
+        # in a new DataLoader and automate the offset based on
+        # DataAttrs like 'inst' for example
+        da = self._maybe_apply_time_shift(da, time_offset, **DataAttrs)
+        start_date_xarray = times.numpy_datetime_range_workaround(start_date)
+        end_date_xarray = start_date_xarray + (end_date - start_date)
+        return times.sel_time(da, np.datetime64(start_date_xarray),
+                              np.datetime64(end_date_xarray)).load()
+
+    @staticmethod
```
spencerahill, line 194: Moving forward let's make module-level functions instead of static and class methods (unless there's a compelling reason to do otherwise).

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        Parameters
+        ----------
+        file_set : list
+            List of paths to files
+
+        Returns
+        -------
+        Dataset
+        """
+        io.dmget(file_set)
```
spencerahill: I would like to avoid hardcoding in this GFDL-specific logic.  For now, can you just wrap this in another function called something like `apply_preload_user_commands` (that name could be better).  e.g.
```
def apply_preload_user_commands(file_set, cmd=io.dmget):
    """Call desired functions on file list before loading.

    For example, on the NOAA Geophysical Fluid Dynamics Laboratory computational cluster, data that is saved on their tape archive must be accessed via a `dmget` (or `hsmget`) command before being used.
    """
    if cmd is not None:
        cmd(file_set)
```
That way it will be easier to later enable users to specify this with their own needed system calls.

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        return xr.open_mfdataset(file_set, preprocess=cls.rename_grid_attrs,
+                                 concat_dim=internal_names.TIME_STR,
+                                 decode_cf=False)
+
+    @staticmethod
+    def _prep_time_data(ds):
+        """Prepare time coordinate information in Dataset for use in
```
spencerahill: (here and elsewhere) 1 line summary to start docstring

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        Dataset
+            Dataset with grid attributes set as coordinates
+        """
+        int_names = internal_names.GRID_ATTRS.keys()
+        grid_attrs_in_ds = set(int_names).intersection(set(ds.coords) |
+                                                       set(ds.data_vars))
+        ds.set_coords(grid_attrs_in_ds, inplace=True)
+        return ds
+
+    def _generate_file_set(self, var=None, start_date=None, end_date=None,
```
spencerahill, line 203: Since the call signature differs so much among the different DataLoaders, does it make sense to have this call signature just be (self, *args, **kwargs)?  It's confusing in e.g. the DictDataLoader to have all of these named but unused kwargs.

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        try:
+            return self.file_map[intvl_in]
+        except KeyError:
+            raise KeyError('File set does not exist for the specified'
+                           ' intvl_in {0}'.format(intvl_in))
+
+
+class NestedDictDataLoader(DataLoader):
+    """A data loader that locates files based on a nested dictionary mapping of
+    intvl_in to dictionaries mapping variable names to lists of files.
```
spencerahill: describe intvl_in in words

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+                return self.file_map[intvl_in][name]
+            except KeyError:
+                pass
+        raise KeyError('Files for the var {0} cannot be found in for the '
+                       'intvl_in {1} in this'
+                       ' OneDirDataLoader'.format(var, intvl_in))
+
+
+class GFDLDataLoader(DataLoader):
+    """A data loader that locates files based on GFDL post-processing naming
```
spencerahill: One-line summary: "Data Loader for NOAA GFDL model output"

And afterwards define GFDL (=NOAA Geophysical Fluid Dynamics Laboratory)

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+            else:
+                self.data_end_date = template.data_end_date
+        else:
+            self.data_direc = data_direc
+            self.data_dur = data_dur
+            self.data_start_date = data_start_date
+            self.data_end_date = data_end_date
+
+    @staticmethod
+    def _maybe_apply_time_shift(da, time_offset=None, **DataAttrs):
```
spencerahill, line 357: Put the old explanatory text re: the offset of GFDL inst data in this docstring.  Otherwise hard to understand L362-373

aospy/data_loader.py
```diff
@@ -0,0 +1,416 @@
...
+        """Special logic to aid in automation of time offsetting for GFDL
+        post-processed data"""
+        if time_offset is not None:
+            time = times.apply_time_offset(da[internal_names.TIME_STR],
+                                           **time_offset)
+            da[internal_names.TIME_STR] = time
+        else:
+            try:
```
spencerahill: Why is this try/except guard needed?

spencerahill (commented):
This is great work overall.  Most of my comments are stylistic/nits.  I hope they don't read as blunt.

We ultimately need users to be able to define their own data loaders in addition to using the ones that we pre-package aospy with.  Is that currently possible with the current DataLoader API?  I.e. if they define their data loader class in e.g. their aospy object library, can they use it?  If yes, we should document that (although not necessarily in this PR).

spencerkclark:
@spencerahill many thanks for taking a comprehensive look at this.  I'll try and work through your comments / questions tomorrow.

Regarding your broad question:
> We ultimately need users to be able to define their own data loaders in addition to using the ones that we pre-package aospy with. Is that currently possible with the current DataLoader API? I.e. if they define their data loader class in e.g. their aospy object library, can they use it? If yes, we should document that (although not necessarily in this PR).

The answer right now is yes, but with a caveat.  The caveat is that the custom DataLoader's ```_generate_file_set``` method must depend on arguments currently taken in the Calc object constructor.  In other words, the parameters needed to identify a file set to use for a given calculation must be in the set: ```var```, ```start_date```, ```end_date```, ```domain```, ```intvl_in```, ```dtype_in_vert```, ```dtype_in_time```, ```intvl_out```.

At some point we need to address this issue.  Since these parameters also play a role in constructing the filenames of aospy output, this becomes a little complicated (and moves beyond the realm of just loading data).  DataLoaders have done a decent job now of cleaning up Model and Run constructors (so that they don't need to have long lists of arguments that may or may not be used depending on the situation), but this has not solved that same problem in Calc.  

Though maybe there's a simple solution that I'm not seeing.  Does that make sense?  Should I open an issue to track further discussion?

spencerahill:
I see what you mean.  This will require a little thinking to get right, so yes please open an issue.

aospy/model.py
spencerkclark: The issue is that not all DataLoaders necessarily need ```data_start_date``` and ```data_end_date``` attributes.  On revisiting this it occurs to me one can obtain the desired behavior here more elegantly (and transparently) with ```getattr```, with a default value of ```None```.

aospy/run.py
spencerkclark: Addressed the same way as above:

```python
self.default_start_date = datetime_or_default(
    default_start_date, getattr(data_loader, 'data_start_date', None))
self.default_end_date = datetime_or_default(
    default_end_date, getattr(data_loader, 'data_end_date', None))
```

aospy/test/data/objects/examples.py
spencerkclark: Sorry this was buried in the commit message of 36b526c.  They have been changed (only to add the ```time_bounds``` variable):

> Initial steps towards integrating DataLoader into Calc  …
This is far from finished, but the limited tests we have are passing. Note that I modified the example .nc files to include the CF-required time_bounds variable.

aospy/__config__.py
spencerkclark: Agreed

aospy/__init__.py
spencerkclark: Addressed in a677b93

aospy/calc.py
spencerkclark: Addressed in a677b93

aospy/internal_names.py
spencerkclark: Addressed in a677b93

I ended up being a little more careful about the distinction between variables and coordinates here; let me know what you think.

aospy/__init__.py
```diff
@@ -1,6 +1,6 @@
 """aospy: management, and analysis of gridded climate data."""
-from .__config__ import (user_path, LAT_STR, LON_STR, PFULL_STR, PHALF_STR,
-                         PLEVEL_STR, TIME_STR, TIME_STR_IDEALIZED, YEAR_STR)
+from .__config__ import user_path
+from . internal_names import *
```
spencerahill: Let's omit this.  I think it's fine to access the names through internal_names, and wildcard imports are generally frowned upon.

Or if you have a reason to keep them, change to explicit imports.

spencerahill:
@spencerkclark ping me whenever you're done with this round of edits (no rush)

aospy/test/test_data_loader.py
spencerkclark: Yes, I am wondering why ```_maybe_apply_time_shift``` modifies its DataArray argument in place.  Ideally I wouldn't want it to.

aospy/test/test_data_loader.py
spencerkclark: This is reflective of my ignorance of all the possible use cases enabled by the GFDLDataLoader (I've only used it a few times a while back).  Could you point me to an example for when ```'monthly_from_'``` is in ```dtype_in_time```?  I was honestly just guessing here.

aospy/test/test_data_loader.py
spencerkclark: I should have been more descriptive here.  Within ```utils.io.data_name_gfdl``` it is stated:
```python
elif intvl_type in ['seasonal', 'seas']:
      # 2015-04-29: This function is obsolete.  Should fix this
      # eventually.  But I almost never use seasonal data, so not a
      # priority.
      # label, val = _intvl_indices_and_label(intvl, intvl_type)
      label = label.upper()
```
The way the function is written currently, you'll get a ```NameError``` whenever this ```elif``` statement is active, because ```label``` will not be defined before the ```label = label.upper()``` line.  As far as I can tell the function used in the call that defines ```label```, which was commented out here, has been removed from aospy at some point in the past.  Should we just delete this part of the function / tests of it for now?

aospy/data_loader.py
spencerkclark: Good call, this is not needed

aospy/test/data/objects/examples.py
spencerkclark: Yes, you can, very nice!

aospy/data_loader.py
spencerkclark: This is related to the issue I referenced last night in Calc.  Since at the moment Calc always provides all of those arguments, we can't limit the arguments passed to ```_generate_file_set```, otherwise we'll get a ```TypeError``` due to the specification of additional, undefined keyword arguments.  So for now I think it might be best to be explicit here (since we can't shorten the call signatures of the overriding versions).

spencerkclark:
@spencerahill I *think* I just finished going through everything.  Let me know your thoughts.  Thanks!

spencerahill:
@spencerkclark thanks!  Tied up until later tonight (>8pm PST); will take another look then and will merge if no major problems.

spencerkclark:
Sure thing -- there are a few lingering (relatively minor) questions I've interspersed throughout your review comments that might prompt a few more edits.  I'll be offline after 10 PM eastern time, so any conversation that happens after that I won't get to until tomorrow morning (so I'm perfectly fine if we don't merge this tonight).

spencerkclark:
@spencerahill somewhat of an aesthetic question -- what do you think of the name "DataLoader"?  Maybe I've been looking at it too much lately, but it feels a little awkward to me.  

What do think about "DataSource"?  I feel like that might be more evocative of its primary purpose (to the user): to define where data files are located, and how to find them.  For the most part in aospy, the user doesn't actually think about loading the data, but they definitely need to think about where their data files are, and point aospy in that direction in their object libraries.

aospy/test/data/objects/examples.py
spencerahill: Cool, thanks.  Should have actually read the commit messages!

spencerahill:
>what do you think of the name "DataLoader"? 

Let's leave it for now and see how it feels after a while.  I'm not opposed in principle to changing the name, but DataSource doesn't seem sufficiently better to me right now.

aospy/test/test_data_loader.py
spencerahill: Ha I have no recollection of ever seeing that either.  So let's get rid of it.

aospy/test/test_data_loader.py
spencerahill: Gotcha.  This works and the test passes by just changing this to `label = intvl.upper()`.  I don't use these file types often but they are very regularly outputted, so let's go that route.

aospy/data_loader.py
spencerahill: Ha, I see the explanatory text I was referring to doesn't actually exist!  My bad.

spencerahill:
@spencerkclark as a small token of gratitude for your tireless work on this PR, I saved you a few minutes of trouble by committing my final edits myself.

Thank you and congrats!  This is as big a change to the internals as we've ever done, and is a critical step in our ongoing efforts to revamp the codebase.

spencerkclark:
Awesome, thanks again for your feedback and help!

PR #92, opened 2016-10-12
Configure Travis CI and Coveralls

spencerkclark:
@spencerahill This is the setup I used in aospy-db.  Maybe we can give this a try?  (I would squash merge to clean up the commit history).


spencerkclark:
@spencerahill It looks like this attempts to run the test suite, but there are a number of import errors we'll need to address (see the Travis Details link).  I'll dig a little deeper when I get a chance.

spencerkclark:
@spencerahill Looks like it's working now -- I needed to add a bit of hack to prevent an import error from infinite-diff in utils.py (b2d89c7).  I'm not sure if there's a better way of doing that -- maybe it's OK to leave this hack in, but put some logic in the one method that needs CenDiff in utils to raise a meaningful error if it is called and CenDiff is not imported.  What are your thoughts?

spencerahill:
Thanks @spencerkclark, this is great!  I hope you don't mind if I copy this over to infinite-diff, animal-spharm, and my aospy-obj-lib!

Actually, this is making me wonder of we should remove the infinite-diff dependency entirely, and to just hard copy over the needed functionality.  The long-term prospects of infinite-diff are much less solid in my mind than those of aospy.  I've opened an issue about this: #93

> maybe it's OK to leave this hack in, but put some logic in the one method that needs CenDiff in utils to raise a meaningful error if it is called and CenDiff is not imported.

Yes, if you wouldn't mind adding that to this PR, I'd say that's a totally fine solution for the time being.

spencerkclark:
@spencerahill By all means, feel free to copy things over to your other libraries!

That sounds good -- I just added the updated error logic to `utils.py`.  If that checks out I think we can go ahead and squash merge.

spencerkclark:
@spencerahill Also, just so you're aware, I turned off the Coveralls comment-bot, and email notifications from Travis.  Let me know if you'd like either of those turned back on.

spencerahill:
> I turned off the Coveralls comment-bot, and email notifications from Travis.

Let's leave the coveralls bot on for now.  If it gets annoying we can turn it off later.

I'm struggling to interpret the Coveralls report: https://coveralls.io/builds/8320927.  For example, timedate.py obviously is tested in places, but the coverage report says that it has 0 lines covered.  Likewise for the other main modules -- only the test_\* files themselves and the test_objs/ files show nonzero coverage.

spencerahill:
@spencerkclark once that's sorted out this is ready to merge.

spencerkclark:
Yeah I'm confused by that as well.  Perhaps the comment on [this SO question](http://stackoverflow.com/questions/28289672/coveralls-shows-0-coverage-for-node-js-project) is relevant?

spencerkclark:
@spencerahill This is definitely on our end (not coveralls) -- I'm able to reproduce the prior false coverage reports locally.  I _think_ things are working as we'd expect now, but I can't say I have a good understanding for why swapping  `pip install -e .` for `python setup.py install` fixes things.

spencerahill:
@spencerkclark thanks for digging into this.  Sorry, you were right -- let's turn back off the Coveralls bot.  Distracting.

> This is definitely on our end (not coveralls) -- I'm able to reproduce the prior false coverage reports locally. I think things are working as we'd expect now, but I can't say I have a good understanding for why swapping pip install -e . for python setup.py install fixes things.

Hmm, that is definitely odd.  But if you think it does the job and doesn't have any unintended side-effects, I'd say let's call this done, and then open a new issue if it becomes a problem again.  Sound OK?

spencerkclark:
Sounds good, I just turned off the bot.  

> Hmm, that is definitely odd. But if you think it does the job and doesn't have any unintended side-effects, I'd say let's call this done, and then open a new issue if it becomes a problem again. Sound OK

Yeah, I'd like to get to the bottom of this eventually, but as far as I can tell it gets the job done.  Most packages I've seen use `python setup.py install` in that spot in their `.travis.yml`, but some do use `pip install -e .` (e.g. [xgcm](https://github.com/xgcm/xgcm/blob/master/.travis.yml)).  It seems like it's something subtle.

But I agree, I think this is safe to merge for now (I just removed some unnecessary dependencies).

spencerahill:
Great, thanks again!  Hoping to find time this week or next to get the doc builds working to our liking (part way there but still a couple snags), at which point we'll be in pretty good shape as regards these non-core-code aspects of the package.

PR #95, opened 2016-10-14
WIP: Add infrastructure for sphinx-based documentation

spencerahill:
Initial stab at automatically generating docs using [sphinx](http://www.sphinx-doc.org/en/stable/index.html).

The resulting docs have a long way to go in order to be actually useful, but I think this is a good start for now.  I would like to merge this once we're sure there's no glaring problems; then we can incrementally improve on the documentation.

I'm also in a hurry to merge because, as far as I can tell, RTD won't update the build until we have merged: http://aospy.readthedocs.io/en/latest/

@spencerkclark can you give this a look-over and generate the HTML on your local machine?  You'll need sphinx, which you can get via conda.  Then just run `make html` from `aospy/docs`.  If there's no bugs (and you don't have any other major objections), I'd say we're good to go.

After the merge I'll make a new issue with a checklist of docs-related todo's.


spencerahill:
You'll also need the RTD sphinx theme I'm using: `conda install sphinx_rtd_theme`

docs/intro.rst
```diff
@@ -0,0 +1,11 @@
+Introduction
+============
+
+aospy is a framework for automating the analysis of climate model output and the management of the resulting data.  This enables automated executiong across an arbitrary number of models, simulations, calculations, time periods, and other parameter settings -- even if the internal variable names or other settings are not consistent across models or simulations.
```
spencerkclark: Minor typo: "This enables automated _executiong_ ..."

spencerkclark:
This looks great to me!  I'm able to build locally with no issues.  Thanks for setting this up.

spencerahill:
Great, thanks!

Issue #94, opened 2016-10-13
Restore original datetime metadata upon saving output


spencerkclark:
Currently, if we encounter time dimensions referenced to a date prior to 1678-01-01, we automatically offset things internally such that the dates aospy sees all begin at 1900-01-01 (this is to account for restrictions in `pandas`).  When we save computational results to netCDF files, however, we do not currently restore datetime metadata to match the way things were in the initial run (so output netCDF files look as though things started in 1900 as well).  This may amount to simply saving the initial units attribute of the time variable when reading in the file, and resetting it to the initial value when we save data.

In addressing this issue I think we should also consider changing the offset reference date to 1678-01-01 to allow the maximum available `pandas` / `np.datetime64` compatible time range.

spencerahill:
> In addressing this issue I think we should also consider changing the offset reference date to 1678-01-01 to allow the maximum available pandas / np.datetime64 compatible time range.

This is addressed in #97.

But otherwise yes this would be good to implement and I doubt too difficult.  Although since it's not urgent, maybe it's better to do as part of the Calc refactor, whenever we get to that portion of Calc.

spencerahill:
@spencerkclark is this in place yet or no?  Please close if yes.

spencerahill:
@spencerkclark where do we currently stand on this?  (I'm in the process of culling obsolete Issues as part of #206)

spencerkclark:
@spencerahill it has not been resolved; it's something that would be nice to do, given our current workaround for dates pre-1678, but not necessarily super high priority.  (It might be a low-hanging fruit though).

spencerahill:
Cool, thanks.  `subset_` and `raw_data_{start,end}_date` are written to output netCDF files currently, but these alone can't be used to infer the original dates.  Still I agree, should be relatively straightforward.

Issue #34, opened 2015-11-07
Create base class that `Proj`, `Model`, and `Run` all inherit from


spencerahill:
The data structures `Proj`, `Model`, and `Run` don't obey an "is-a" conceptual relationship relative to one another (a model is not a type of project, and a run is not a type of model), and thus shouldn't inherit from each other.  It's more of a "has-a" relationship: a project has one or more model; a model has one or more runs).  

Nevertheless, each of them can be conceptualized as being a particular kind of container of data describing climate model/obs data saved somewhere on disk.  So there should be a parent class, something like `InputDataSpecs`, that they each inherit from.  Since we are wanting all specifications about the data to be made at any of the levels, c.f. #28 #27 and #26, this could be implemented all within this parent class, with any level-specific variations being done through overriding/adding methods within that particular child level.  This may ultimately render the `Proj`, etc. classes with very little unique to them, but they should still be retained to facilitate easy organization/calculation specifications.

`Var` is a little trickier, because in addition to specifying _where_ the data can be found (although its not really used in this way in my workflow at least), it also specifies physical aspects of the quantity being represented.  So this requires more thinking -- one option would be to have two classes, one for each of these functions, with the former also inheriting from this parent class.

spencerkclark:
This sounds like a solid strategy for how to consolidate code for like functionality in `Proj`, `Model`, and `Run`.  

With regard to `Var` could you clarify what you mean by "specifying where the data can be found?"  I don't believe I use it this way in my workflow either.  Is there a reason to support such functionality?

spencerahill:
Sure.  I mean that we should support mappings of variables to the external data at the level of a `Var` (in which case obviously it would be of that one variable).  For example, suppose because of a postprocessing error, all of the variables can be found using a simple mapping except for one.  Then, the user could create a `Var` object with the mapping unique to that variable's data.

Now that I've written that out, I'm convinced that this type of functionality should be separate from what `Var` does in its current form, namely specify physical/functional parameters about a particular physical quantity.  Maybe keep `Var` intact and create `VarData` for the data-finding stuff.

spencerkclark:
Cool, that makes sense.  I hadn't thought about exceptions to implicit file mappings.  I agree about keeping this outside the current `Var` class.  Perhaps we could just have an argument in creating a FileMapGenerator that allowed the user to add custom mappings for specific variables that would override the implicit mapping?

spencerahill:
[...]

PR #97, opened 2016-10-24
WIP refactor timedate

### LINKED ISSUE #96: timedate.apply_year_offset causes datetime overflow for sufficiently large start year
Due to the nanosecond resolution, the lower and upper bounds on dates are 1677-09-21 and 2262-04-11 (e.g. [here](http://pandas-docs.github.io/pandas-docs-travis/timeseries.html#timestamp-limitations)).  `timedate.apply_year_offset` adds the constant `TimeManager.YEAROFFSET`, which is currently set to 1899, if the inputted `datetime.datetime` object starts before the minimum year.

This causes an overflow and a Pandas `OutOfBoundsDatetime` error if any of the inputted years are greater than 2262-1899=363, because then the start year exceeds the 2262 bound:

``` python
In [1]: import aospy
import
In [2]: import datetime

In [3]: aospy.TimeManager.apply_year_offset(datetime.datetime(363,1,1))
Out[3]: Timestamp('2262-01-01 00:00:00')

In [4]: aospy.TimeManager.apply_year_offset(datetime.datetime(364,1,1))
/Users/spencer/miniconda3/lib/python3.5/site-packages/pandas/tseries/tools.py in _convert_listlike(arg, box, format, name)
    408             try:
--> 409                 values, tz = tslib.datetime_to_datetime64(arg)
    410                 return DatetimeIndex._simple_new(values, name=name, tz=tz)

pandas/tslib.pyx in pandas.tslib.datetime_to_datetime64 (pandas/tslib.c:29768)()

TypeError: Unrecognized value type: <class 'datetime.date'>

During handling of the above exception, another exception occurred:

OutOfBoundsDatetime                       Traceback (most recent call last)
<ipython-input-4-17b710b49c96> in <module>()
----> 1 aospy.TimeManager.apply_year_offset(datetime.datetime(364,1,1))

/Users/spencer/Dropbox/py/aospy/aospy/timedate.py in apply_year_offset(cls, date)
    126             offset = 0
    127         return pd.to_datetime(cls.ymd_to_numpy(date.year + offset,
--> 128                                                date.month, date.day))
    129
    130

/Users/spencer/miniconda3/lib/python3.5/site-packages/pandas/util/decorators.py in wrapper(*args, **kwargs)
     89                 else:
     90                     kwargs[new_arg_name] = new_arg_value
---> 91             return func(*args, **kwargs)
     92         return wrapper
     93     return _deprecate_kwarg

/Users/spencer/miniconda3/lib/python3.5/site-packages/pandas/tseries/tools.py in to_datetime(arg, errors, dayfirst, yearfirst, utc, box, format, exact, coerce, unit, infer_datetime_format)
    289                         yearfirst=yearfirst,
    290                         utc=utc, box=box, format=format, exact=exact,
--> 291                         unit=unit, infer_datetime_format=infer_datetime_format)
    292
    293

/Users/spencer/miniconda3/lib/python3.5/site-packages/pandas/tseries/tools.py in _to_datetime(arg, errors, dayfirst, yearfirst, utc, box, format, exact, unit, freq, infer_datetime_format)
    427         return _convert_listlike(arg, box, format)
    428
--> 429     return _convert_listlike(np.array([arg]), box, format)[0]
    430
    431 # mappings for assembling units

/Users/spencer/miniconda3/lib/python3.5/site-packages/pandas/tseries/tools.py in _convert_listlike(arg, box, format, name)
    410                 return DatetimeIndex._simple_new(values, name=name, tz=tz)
    411             except (ValueError, TypeError):
--> 412                 raise e
    413
    414     if arg is None:

/Users/spencer/miniconda3/lib/python3.5/site-packages/pandas/tseries/tools.py in _convert_listlike(arg, box, format, name)
    396                     yearfirst=yearfirst,
    397                     freq=freq,
--> 398                     require_iso8601=require_iso8601
    399                 )
    400

pandas/tslib.pyx in pandas.tslib.array_to_datetime (pandas/tslib.c:41972)()

pandas/tslib.pyx in pandas.tslib.array_to_datetime (pandas/tslib.c:40843)()

pandas/tslib.pyx in pandas.tslib.array_to_datetime (pandas/tslib.c:39338)()

pandas/tslib.pyx in pandas.tslib.array_to_datetime (pandas/tslib.c:39232)()

pandas/tslib.pyx in pandas.tslib._check_dts_bounds (pandas/tslib.c:29245)()

OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 2263-01-01 00:00:00
```

@spencerkclark, what if, rather than applying a constant offset, we just directly set the start year to be 1678?  The catch is then retaining the original year values.  But looking through the code and remembering our most recent conversation (although I can't find it on GH), right now we aren't even bothering to change the years back anyways...am I remembering that right?

spencerahill:
Fixes #96.

In working on this, I realized that much of timedate's logic was needlessly convoluted, and in some cases we were just making crummier versions of existing pandas and datetime functionality.  So I have started in on cleaning up that module.  But I think more should be done.

In particular, the `TimeManager` class was almost entirely static or class methods.  Having moved them to be module-level functions, I'm now wondering if there's any need for the TimeManager class at all.

Also, `to_datetime` feels hack-ish.  My sense is that if we were more careful with our logic in `calc` (and maybe elsewhere), we wouldn't need this utility function that catches all these different types.  At the very least, catching bools and returning them seems weird.

Even having pruned a fair bit, it still feels like there's room to clean things up further.  But I need to move on to other things for the time being.

TODOs:
- [x] Make decisions on the above considerations.
- [x] More tests.
- [x] Further improve docstrings.


spencerkclark:
> In particular, the TimeManager class was almost entirely static or class methods. Having moved them to be module-level functions, I'm now wondering if there's any need for the TimeManager class at all.

My inclination is to remove the TimeManager class as well.  I wonder though if there is value in having something external to Calc to store the original time metadata, for reference later (i.e. to address #94)?  Then again, if an upstream solution surfaces, this will no longer be an issue (see #98 for more discussion).

> Also, to_datetime feels hack-ish. My sense is that if we were more careful with our logic in calc (and maybe elsewhere), we wouldn't need this utility function that catches all these different types. At the very least, catching bools and returning them seems weird.

Very much agreed.  My impression is the only use case for this function is accepting multiple different ways of representing dates as input parameters (e.g. start and end dates for calculations).  If we unilaterally say we'll only support certain ways of inputting dates, then we won't need this messy logic to interpret user input.  Is that right?

spencerahill:
> Then again, if an upstream solution surfaces, this will no longer be an issue (see #98 for more discussion).

Exactly.  

> My inclination is to remove the TimeManager class as well. I wonder though if there is value in having something external to Calc to store the original time metadata, for reference later (i.e. to address #94)?

I think this matters to the extent that, following the Calc refactor, there are 1 or more non-Calc objects in which time arrays are stored, which I do think will be the case.  (Funny enough, currently the `self.date_range` attribute generated in `calc.py` via `tm.create_time_array()` is never used, at least elsewhere in `calc.py`.  

> If we unilaterally say we'll only support certain ways of inputting dates, then we won't need this messy logic to interpret user input. Is that right?

Exactly.  This ultimately stems from my own ineptitude when I originally wrote this -- at the time, I didn't understand the importance of interfaces, and that your classes should be able to expect  a particular input type and do something with it, not deal with inferring from lots of different input types what to do.  That goes upstream.

Also at this point, I'm not even sure if the `bool` case ever occurs -- will look into it.

aospy/timedate.py
```diff
@@ -8,113 +8,63 @@
...
-            end=self.apply_year_offset(self.end_date), freq='M'
-        )
-        time = xr.DataArray(all_months, dims=[TIME_STR])
-        month_cond = self._construct_month_conditional(time, self.months)
-        return time[month_cond]
+    """
+    if isinstance(obj, (bool, datetime.datetime)):
+        return obj
+    if isinstance(obj, str):
+        return pd.to_datetime(obj).to_datetime()
```
spencerkclark: I think the way things are set up now, this won't work if one specifies a date as a string that cannot be converted to a `Timestamp` object due to range issues.  Is that right?  

If we ultimately want plain `datetime.datetime` objects, then maybe we should use the [datetime.datetime.strptime](https://docs.python.org/2/library/datetime.html#datetime.datetime.strptime) function directly?  E.g.

``` python
In [1]: from datetime import datetime

In [2]: datetime.strptime('2000-01-01', '%Y-%m-%d')
Out[2]: datetime.datetime(2000, 1, 1, 0, 0)

In [3]: datetime.strptime('0001-01-01', '%Y-%m-%d')
Out[3]: datetime.datetime(1, 1, 1, 0, 0)
```

spencerahill: Yes, good catch.  `strptime` may work, but then we are locked into the '%Y-%m-%d' format.

 Currently `to_timedate` is called by `Run.__init__` on `data_in_{start,end}_time` and on `start_date` and `end_date` of Calc.  As such, we would need to check/require in these places that the strings follow our desired format.

Another option is doing away with string-specified dates altogether, and thus avoiding trying to parse strings entirely.  So users would be able to specify dates either as an integer or as a datetime-like object.

spencerahill: > Another option is doing away with string-specified dates altogether, and thus avoiding trying to parse strings entirely.  So users would be able to specify dates either as an integer or as a datetime-like object

The more I think about this, the more I like it.

aospy/timedate.py
```diff
@@ -8,113 +8,63 @@
...
+    if isinstance(obj, int):
+        if end_of_year:
```
spencerahill: @spencerkclark, would this logic fail for e.g. 360 day calendar?  Since 12-31 isn't a valid date.

aospy/timedate.py
spencerkclark: Yeah actually I would lean towards getting rid of string-specified dates as well.  What types of datetime objects should we accept as input? 

xarray accepts `pd.Timestamp`, `np.datetime64`, and `datetime.datetime` objects as bounds for time-slices, so long as the dates are in the Timestamp-valid range.  So if we didn't have to worry about the range issue, I think we could accept all three and not even worry about casting to a consistent type.

Coming up with a solution that works with the range issue is a little trickier (since adding an offset to the year might not have the same syntax for the different datetime types); in that case we might need some logic to cast objects to be a certain type first (or accept only `datetime.datetime` objects, since that is what we use to do the year offset).

aospy/timedate.py
spencerkclark: Yes, it looks like it would; my instinct would be to remove this kind of logic.  While convenient for some cases, I think it may assume too much (in particular that any input of type `int` has units of years).

aospy/timedate.py
```diff
@@ -132,3 +82,67 @@ def _get_time(time, start_date, end_date, months, indices=False):
...
+        raise ValueError(message)
+    st_ind = first_letter.find(months.lower())
+    return np.arange(st_ind, st_ind + len(months)) % 12 + 1
+
+
+def _construct_month_conditional(time, months):
+    """Create a conditional statement for selecting data in a DataArray."""
+    cond = False
+    for month in months:
+        cond |= (time['time.month'] == month)
```
spencerkclark: Should we make `time.month` a global variable, or perhaps write this as:

``` python
cond |= (time['{}.month'.format(TIME_STR)] == month)
```

I think `month` is a property of `pd.DatetimeIndex`, so I wouldn't expect it to vary.

aospy/timedate.py
spencerahill: > since adding an offset to the year might not have the same syntax for the different datetime types

Exactly.  So I am inclined to initially just support a single type, and return to that later if it becomes too restrictive.  I would lean towards `datetime.datetime` since it's what we're using already, but isn't that important to me.

aospy/timedate.py
spencerahill: Yes, good catch.

aospy/timedate.py
spencerahill: Hmmm, so you're thinking drop support for int-specified times also?  That is the one convenience functionality that I was thinking we'd retain.  But maybe you're right, let's just go all the way and only accept a single type for now.  Ditto above comment about returning to later if needed.

spencerkclark: We can discuss it :)

I might be more open to it if we made it more explicit what the integer's units were.  If we did that, I think at the same time it might be worth adding support for integer ranges of days as well (a common analysis practice in idealized models with no seasonal cycle).  Though I agree, in terms of refactoring, it may be simpler to do things one way that covers everything, and then add convenience features later.

spencerahill: In which case there is no need for this function at all, right?  Instead, at the Run and Calc levels directly, we just enforce that the input is of our required datetime type.

spencerkclark: Exactly

spencerahill: Good point re: integers for day ranges.  OK, further pushing me to say we only accept `datetime.datetime` for now...as the zen of python states, "explicit is better than implicit"!

spencerahill:
@spencerkclark I have removed `TimeManager` and `to_datetime` and done some other cleaning up.  I also went ahead and required that `Run` and `Calc` date-related attributes be of type `datetime.datetime` (fair warning, this requires grepping/replacing a lot of lines in your object library).

Still need more tests.  I'm also thinking of renaming timedate to utils_time, or maybe creating a utils package with separate time and other modules.  What do you think about that?  And any other comments I'd appreciate.  I'm especially looking for ways we could simplify this further as well as make names more descriptive.

spencerahill:
The checking I have implemented within `Run.__init__`, I now realize, should be implemented at all levels -- Proj, Model, and Run -- since we allow users to specify date ranges, etc. at any one of them.  Thus, I think it makes sense to go within it's own class or function.  This doesn't suite seem like the best fit for `DataLoader`, because it's really just the specifications for the data.  `DataSpecs`?

However if we go that route, I still would like the Run, etc. __init__ call signatures to be able to specify these directly.  So we would need to catch those and pass them to the `DataSpecs` constructor.  @spencerkclark, I feel like I'm overlapping with #90 here...what is your take?

aospy/run.py
```diff
@@ -1,51 +1,68 @@
...
+    Returns
+    -------
+    `default` if `date` is `None`, otherwise returns the result of
+    `ensure_datetime(date)`
+
+    """
+    if date is None:
+        return default
+    else:
+        return ensure_datetime(date)
```
spencerahill: This is what I meant in terms of special functionality for these type checks.  So maybe this goes in timedate (or utils/time) and then Model and Proj call it too?  The alternative is a separate class, c.f. my previous comment.

spencerahill:
Also ignore the name changes and failing tests for now...I'll fix those

spencerkclark:
@spencerahill apologies for being slow on this -- I'll try and get some substantive comments to you tomorrow.

spencerahill:
@spencerkclark no rush, especially over the holiday week.

spencerkclark:
> I have removed TimeManager and to_datetime and done some other cleaning up. I also went ahead and required that Run and Calc date-related attributes be of type datetime.datetime (fair warning, this requires grepping/replacing a lot of lines in your object library).

Great, this is all for the better, so I'm cool with doing some grepping/replacing.  Eventually once there is an upstream fix to the datetime issue, I think we'll be able to revert back to taking strings as input, but for now I think this is the best solution.

> I'm also thinking of renaming timedate to utils_time, or maybe creating a utils package with separate time and other modules. What do you think about that?

Yes, I think a utils package would be a good idea (I'm assuming you mean as a directory within this repo).  The current ```utils.py``` module is probably getting too broad, so splitting that into sub-modules would be a good way to organize things.  I think the main modules in that package for now would be ```utils_time``` and ```utils_pressure``` (or perhaps to be more general, something along the lines of  ```utils_vert_coords```).  

Maybe we could throw the ```io.py``` module in there as well, but I might wait on that until #90 is finished since I'll be factoring out the code that generates the filenames to fetch GFDL post-processed data (which will probably leave just methods relating to building filenames of aospy-generated data -- so just things relating to "output" rather than "input and output").  That way we'll have a better idea of what to name it based on what it contains.

> The checking I have implemented within Run.__init__, I now realize, should be implemented at all levels -- Proj, Model, and Run -- since we allow users to specify date ranges, etc. at any one of them. Thus, I think it makes sense to go within it's own class or function. This doesn't suite seem like the best fit for DataLoader, because it's really just the specifications for the data. DataSpecs?

> However if we go that route, I still would like the Run, etc. init call signatures to be able to specify these directly. So we would need to catch those and pass them to the DataSpecs constructor. @spencerkclark, I feel like I'm overlapping with #90 here...what is your take?

In #90 we will need to think of some way to identify a list of files based on user-provided parameters (which could include dates for implicit filenames for GFDL post-processed data).  That said, the way I envision ```DataLoader``` being used, the dates will be passed at runtime through ```calc.py```, presumably already of the desired type (checked upstream either on the instantiation of ```Proj```, ```Model```, or ```Run``` objects through the ```default_date_range``` argument or upon creation of a ```Calc``` object), so no type checking should need to be done there.  Therefore, if we do make some sort of object to handle those file-set identifying parameters related to ```DataLoader```, I don't think dates would need to be part of it.  

So my sense is that there isn't too much overlap between here and #90.  I think in this PR it would be OK to just put the date type checking in the utils/time module and call that from the ```Proj```, ```Model```, and ```Run``` constructors in lieu of a more complicated ```DataSpecs``` class.  Though it's completely possible I haven't thought everything through -- other than dates which other parameters do you think would be part of ```DataSpecs```?

spencerahill:
>Maybe we could throw the io.py module in there as well, but I might wait on that until #90 is finished since I'll be factoring out the code that generates the filenames to fetch GFDL post-processed data

Oops, I got carried away and included io in the new utils subpackage, and I didn't do frequent-enough commits to undo it once I realized my mistake.  I hope this doesn't create too many clashes/required manual merges when we merge #90!  My bad.

>The way I envision DataLoader being used, the dates will be passed at runtime through calc.py, presumably already of the desired type (checked upstream either on the instantiation of Proj, Model, or Run objects through the default_date_range argument or upon creation of a Calc object), so no type checking should need to be done there. Therefore, if we do make some sort of object to handle those file-set identifying parameters related to DataLoader, I don't think dates would need to be part of it.

That's perfect.

>Though it's completely possible I haven't thought everything through -- other than dates which other parameters do you think would be part of DataSpecs?

Essentially all of the `data_` attributes.  One argument in favor of a separate class is testing: if I create a single class, I can test it in only one place, rathr than in all three `__init__`s.  I have actually been thinking about something along these lines for a while; see #34.  Let me give it some more thought.

spencerahill:
The commits I just pushed add the datetime checks to both Run and Model but not Proj, and not via a class as I was just arguing in favor of above.

The bigger issue is that the calc tests (now `test_calc_basic`) are failing, because the time subsetting logic is now somehow breaking and returning an empty time array.  I am out of time right now but will return to this.  The problem lies somewhere in or after `Calc._create_input_data_obj`, in particular the 'since 1900...' logic.

spencerkclark:
> Oops, I got carried away and included io in the new utils subpackage, and I didn't do frequent-enough commits to undo it once I realized my mistake. I hope this doesn't create too many clashes/required manual merges when we merge #90! My bad.

This is all good -- if we can merge this into develop (and this seems like it is well on its way) before I work on refactoring that particular code in #90, it shouldn't be too hard to rebase and go from there.

> Essentially all of the data_ attributes. One argument in favor of a separate class is testing: if I create a single class, I can test it in only one place, rathr than in all three __init__s. I have actually been thinking about something along these lines for a while; see #34. Let me give it some more thought.

Ah, thanks for the clarification.  I envision that the setup of the ```data_``` attributes could also be modified by #90.  My goal is to have those attributes (only where needed) specified in the specific constructors of the various DataLoaders.  For instance, currently one *could* supply a ```data_start_date``` attribute on a Run that uses the ```'one_dir'``` file reference method, but that attribute would never be used (because it is only used by the ```'gfdl'``` file reference method).  I think we should strive to address that in #90.

Ideally, I think it makes the most sense to have DataLoader-specific attributes handled at the level of the different individual DataLoaders.  But, I suppose this does raise a bit a challenge if we wanted certain attributes to be inherited from parent objects, but wanted to override others.  Ultimately, though, I think we could find a way to have the best of both those worlds. 

One idea would be to create a method within DataLoader that would return a copy, with attributes modified via present keyword arguments.  Then one could write something like (note this is not a complete example in terms of full specification of attributes, but hopefully it illustrates the behavior I'm getting at):

```python
am2 = Model(
    name='am2',
    data_loader=GFDLDataLoader(
        data_start_date=datetime.datetime(2000, 1, 1),
        data_end_date=datetime.datetime(2099, 12, 31),
        data_in_dir='path/to/data',
    )
)

control = Run(
    name='control',
    data_loader=am2.data_loader.override(data_in_dir='/path/to/data/control')
)
```
So in this case the ```control``` Run would inherit the ```data_start_date``` and ```data_end_date``` attributes from the ```am2``` Model's DataLoader, but modify the ```data_in_dir``` attribute.  

Though, as I think about this now, this would cause a circular dependency (i.e. right now we need to specify a list of Run objects in a Model constructor, so one couldn't explicitly reference a Model object in a Run constructor as I've done here).  In theory one could fix this by reversing the way the inheritance links were defined and specify the Model a Run is associated with in the Run constructor, but it's possible that could have unintended consequences I haven't thought about.

I'm not sure if that clarifies or clouds your picture :)  I think unifying the common aspects of ```Proj```, ```Model```, and ```Run``` in a single class is a really good idea (as you first suggested in #34), but I think we need to hash out how we want those objects to behave and interact a little more.

spencerahill:
>This is all good -- if we can merge this into develop (and this seems like it is well on its way) before I work on refactoring that particular code in #90, it shouldn't be too hard to rebase and go from there.

Awesome, glad to hear it.

>One idea would be to create a method within DataLoader that would return a copy, with attributes modified via present keyword arguments.

The alternative would be to keep the attributes as kwargs in `__init__` and then use those to constract the DataLoader.  So your example would become:

```python
am2 = Model(
    name='am2',
    data_loader='gfdl' # or GFDLDataLoader
    data_start_date=datetime.datetime(2000, 1, 1),
    data_end_date=datetime.datetime(2099, 12, 31),
    data_in_dir='path/to/data',
    )
)

control = Run(
    name='control',
    parent=am2,
    data_in_dir='/path/to/data/control'
)
```
`Model.__init__` would then use the `data_` attributes to instantiate the DataLoader object for am2, and then Run.__init__ would use its `data_in_dir` but otherwise use am2's data_* attributes to instantiate its DataLoader.

>Though, as I think about this now, this would cause a circular dependency (i.e. right now we need to specify a list of Run objects in a Model constructor, so one couldn't explicitly reference a Model object in a Run constructor as I've done here). In theory one could fix this by reversing the way the inheritance links were defined and specify the Model a Run is associated with in the Run constructor, but it's possible that could have unintended consequences I haven't thought about.

This is a good point.  Ultimately, the point of the hierarchical structure is for bottom-up finding of attributes, i.e. looking at the bottom (Run) for an attribute, and if it's not there moving to the next level (Model), and repeat up to Proj if necessary.  In that sense, it's more logically consistent to specify within a Run instance what it's parent is, not in a Model what its children are.  In fact, although it might be nice for a user to be able to easily see all of a Proj or Model's child objects, the internal logic should otherwise be totally agnostic to an object's children.

I'm sure doing this would break some things, and of course cause you and I some non-trivial modification of our object libraries, but maybe it is worth it.  Need to think more about.

spencerahill:
The latest commit fixes the bugs in calc that I mentioned above.  More tests are needed in utils.times, but once those are in I'd say this PR has already grown a lot in scope and so we should merge.  Especially to the extent (even if just a little) it will trip up #90.

Comments/suggestions/questions still welcome though of course

spencerkclark:
>The latest commit fixes the bugs in calc that I mentioned above. More tests are needed in utils.times, but once those are in I'd say this PR has already grown a lot in scope and so we should merge. Especially to the extent (even if just a little) it will trip up #90.

Yes this PR is looking very good.  I agree, once those last couple tests are written to fill out the test coverage of ```utils.times``` I think we can merge.  When I get a chance, I'll move our DataLoader-related discussion over to #90.

spencerahill:
I rearranged some of the tests, which in hindsight messed up the diffs.  My mistake; there was no real need for the re-arranging.

PR #101, opened 2016-10-28
WIP configuration of Sphinx/RTD

### LINKED ISSUE #93: MAINT: potentially remove infinite-diff dependency
C.f. https://github.com/spencerahill/aospy/pull/92#issuecomment-253494151

This seems like a wait-and-see issue for now.  It is possible that [infinite-diff](https://github.com/spencerahill/infinite-diff) ends up becoming popular/continuing to be useful such that I/somebody continues to maintain it, puts it on conda, etc.  In that case, the dependency is fine to have.

But the opposite case is also possible: I find something better and stop maintaining it.  If that happens, we should remove the aospy dependency on infinite-diff and just hardcode in the needed finite differencing logic (or switch to whatever infinite-diff has been surpassed by).

For now, we have logic via #92  to ignore this import as needed for our tests to pass...that should be modified too once we make a decision one way or the other.

spencerahill:
I think I have resolved the issues such that sphinx.autodoc works on readthedocs.org, as evidenced by a passing build with no obvious error messages (prior builds had many): https://readthedocs.org/projects/aospy/builds/4576219/

And the docs seemed to render on RTD with the autodoc parts incorporated: http://aospy.readthedocs.io/en/latest/aospy.html#

However, I'm getting weird "DNS server not found" errors intermittently when I go to our docs.  Maybe it's a local or temporary thing? @spencerkclark could you take a look?

Also, I'm getting a weird Travis CI error due to our basemap dependency.  On my laptop, I can `pip install basemap` just fine, but on Travis pip is complaining that it can't find it.

I would like to merge once these two issues are resolved.  This also closes #93, since the infinite diff dependency (which is not on PyPI/conda) was also throwing error messages.


spencerkclark:
@spencerahill that's great that the docs seem to be working now!  

Did you merge with the latest version of the develop branch before making changes (i.e. post #92)?  My guess is that's what's causing the Travis issues (for instance I don't see a ci directory on this branch).  We use conda to set up the environment on Travis now (so there should be no need to do a pip install of basemap).

#92 also temporarily addressed the infinite-diff dependency issue.

spencerahill:
@spencerkclark good call!  Forgot that I had branched this off prior to those.  Following a manual merge, Travis and RTD are both working.

After some additional thought, I decided it's cleaner to just hardcode in the centered differencing and remove the infinite_diff dependency.  But your temporary workaround was still useful!

@spencerkclark, can you give this one final look over before I squash and merge?

spencerkclark:
@spencerahill I'm glad that sorted things out!  That was a good idea to just hardcode in the centered differencing.  

Just one final thing before we merge this -- I think we can get away with deleting the requirements.txt file; try deleting it and see if everything still works.  (My fault for not addressing that in #92).

spencerahill:
Nice!  I noticed before that xarray doesn't have one.  Travis and RTD work without it, so we're all set.

spencerahill:
(FYI Accidentally hit enter while trying to edit the commit message; hence "onfiguration" and all the extraneous details.  My bad!)

PR #102, opened 2016-10-28
A quick fix to make dmget optional

spencerkclark:
We will need to put more thought into this eventually on how to support different IO sequences for different modeling centers, but for now this should prevent errors due to a lack of a `dmget` command.


spencerahill:
Perfect, exactly what we need for now.  Thanks @spencerkclark !

Issue #103, opened 2016-10-28
Proj.direc_out is unused


spencerkclark:
Currently there is no way for the user to specify an output directory for aospy data.  Instead we default to the GFDL scratch or archive locations.  How should we approach this?

The fastest fix is to just override the scratch directory if the user specifies a `direc_out` in a `Proj` constructor (see the [direc-out branch on my fork](https://github.com/spencerkclark/aospy/blob/direc-out/aospy/calc.py#L180-L188)), but on thinking about it more I'm not sure if that's a safe solution.  If someone is using aospy on a machine where they have root privileges (e.g. a personal laptop), and they neglect to specify a `direc_out` for their project, would things still run, and a `/work/$USER/` directory be created at the root level?  I don't think this is our intended behavior.  Perhaps a sensible default would be the current working directory?

spencerahill:
Good catch.

The "scratch" v. "archive" nomenclature also isn't really universal: for a personal computer, there's only one filesystem, and neither of these titles is a good fit.

So, this would be expanding the modification appreciably, but what if we switch to "primary" and "secondary" as the labels?  The `direc_out` attribute is then mandatory and specifies the root of the primary output directory.  Maybe it can also accept a two-sequence of strings; if the latter, the first is considered the primary and the second the secondary?  It's not quite this simple though, since the logic is different in the scratch and archive cases...

Thoughts?

spencerkclark:
What if we split things into two keyword arguments in the Proj constructor?  Maybe we could use `direc_out` for the directory of the untarred data, and `tar_direc_out` for the directory containing the tar files?

spencerahill:
Is there a way to keep it more general?  Maybe down the line we want to support other write functionality.  I.e. we would have `output_direc` and `output_strategy` (names open to change).  `output_strategy` would specify which output method to use; the default would be what is now "scratch" (not sure what the best name would be...just "standard"?), and for now the only other option would be what is now "archive" (I like "tar" there).

(Also possible I'm getting carried away, given the time crunch of fixing this.  Totally fine with a quick fix and then re-visiting this after the workshop)

spencerkclark:
Yeah I think I'll go with a quick fix for now, if that's OK.  I think generalizing this might fall under the umbrella of the broader refactor of Calc.  

Splitting into two kwargs actually would be pretty painless (and not too breaking a change for both of us, since this gets defined at the Proj level).  I'll put together a quick PR.

spencerahill:
[...]

Issue #50, opened 2015-12-29
Remove obsolete Constant and Units classes


spencerahill:
In particular, multiplying a `Constant` object with an xray object results in a numpy array, rather than an xray object as desired.  This is just one example of multiple problems with the operator overloads for the `Constant` class...I didn't really know what I was doing when I wrote them!  So all of them need to be inspected, likely re-written, and tested.

As a workaround until then, one can just use the `value` attribute, e.g. `c_p.value` rather than just `c_p`.

spencerahill:
These bugs are also causing failing tests.

spencerkclark:
I noticed some remaining messy logic related to constants as I was reviewing #178.  E.g. in [calc.py](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L392-L396):
```python
 # Pass numerical constants as is.
 if isinstance(var, (float, int)):
     return var
 elif isinstance(var, Constant):
     return var.value
```
Returning just the value associated with the constant is needed because the operators (as stated in issue) don't work as desired.  

It occurred to me that what's implemented in [constants.py](https://github.com/spencerahill/aospy/blob/develop/aospy/constants.py) was basically an attempt to create numbers (which you can do mathematical operations with arrays with) with attached metadata (implemented pre-xarray's introduction to aospy).  I *think* the basic functionality desired here could be obtained by simply using single-valued DataArrays with `'units'` and `'description'` attributes.  Then we would get numerical operator support for free.  E.g.
```python
R_d = Constant(
    287.04,
    'J/K/kg',
    description='Dry air gas constant'
)
```
could instead just be:
```python
import xarray as xr

R_d = xr.DataArray(287.04, attrs={'units': 'J/K/kg', 'description': 'Dry air gas constant'})
```

Not a huge priority, but just wanted to get these thoughts down while they were fresh in my head.  Is there any reason to have a separate class for Constant values, or could this DataArray solution suffice?

spencerahill:
>basically an attempt to create numbers (which you can do mathematical operations with arrays with) with attached metadata 

Exactly

>I think the basic functionality desired here could be obtained by simply using single-valued DataArrays with 'units' and 'description' attributes. 

Very good point.  I don't see any reason now or in the future why we the DataArray version wouldn't suffice.  So I'm in favor of dropping the Constants class entirely and replacing the ones we have defined with DataArrays.

[...]

PR #105, opened 2016-10-28
Enable user-specified output directories in Proj

spencerkclark:
This PR provides the quick fix discussed in #103.  It also adds local test data files, which add some basic test coverage to Calc and also verify that output files are saved to the right places.


spencerkclark:
@spencerahill I was a bit clumsy getting the tests configured properly (it took me a while to realize how to set up the paths to the test data on Travis), but I think I've got things working.  Let me know if you're on board with how I made the changes in Calc, and if you're good with the test structure.  Thanks!

spencerkclark:
@spencerahill just so you know, I just rebased / squashed the commits offline to clean up the commit message

aospy/test/test_basic_calc.py
```diff
@@ -0,0 +1,130 @@
...
+from local_test_objs.projects import example
+from local_test_objs.models import im
+from local_test_objs.runs import im_example
+from local_test_objs.variables import condensation_rain, precip
+from local_test_objs.regions import globe, sahel
+
+
+class TestBasicCalc(unittest.TestCase):
+    def setUp(self):
+        print os.getcwd()
```
spencerahill: Did you want to keep this in?  If so, include `from __future__ import print_function` and then use `print(os.getcwd())` for Python 3 compatibility.

spencerahill:
Strange, something with my local setup is causing errors when I try to run the new test (even though they're working on Travis): https://gist.github.com/spencerahill/90e751781812266f21dfdd01d3a4a426

I have seen that `fp` error from scipy before, but can't recall precisely what causes it.  I have confirmed that the test netcdf files are in place and not corrupted.

I'll have to come back to this.

spencerkclark:
@spencerahill How did you go about setting up your environment?  One thing that stands out immediately is that it seems it doesn't have dask installed, which is required for a call that gets made to `xr.open_mfdataset` in `model.py`.  netCDF4 is also a requirement to run the tests now (it wasn't in the past); it is needed for decoding CF attributes.  Check out the new `ci/environement-py27.yml`, for the comprehensive list.

Addressing the import errors may fix everything (the failed tests are simply due to the fact that files are not produced, because of errors upstream -- the import errors could account for all of that since all the calculations start by reading in grid files).  

That said, you seem to be using python 3.5 and until now we didn't have Travis set up to test in that environment.  I added that to our set up, but as you can see, there are issues with imports of the test aospy object libraries within the test modules that prevent the tests even from being fully collected (for python 3.5 but not python 2.7).  I'm not sure what's causing that (in your local environment that doesn't seem to have been a problem).

spencerkclark:
@spencerahill Things seem to be working as we'd expect now for python 3.5; all I needed to do was switch from implicit relative imports to explicit ones in `test_objs/`.  E.g. go from this:

``` python
import models
```

to this:

``` python
from . import models
```

The only thing is that there are now some "unexpected success" tests in the `test_aospy` module associated with division.  How should we handle those?

spencerahill:
> unexpected success" tests in the test_aospy module associated with division.

Let's just remove the decorator.  Probably at some point I screwed up the division logic, saw the tests were failing and slapped it on, and then forgot to change the tests whenever it was fixed again!

spencerahill:
> netCDF4 is also a requirement to run the tests now (it wasn't in the past); it is needed for decoding CF attributes.

Yes, sorry, this was my rookie mistake...borrowing my wife's computer and apparently forgot to install netCDF4.

I'm now getting 100% passing tests on 2.7 and 3.5 with the aforementioned expected failures fix.  That resolved, will finally turn to the actual substance of what you've done!

aospy/calc.py
```diff
@@ -865,19 +865,19 @@ def _update_data_out(self, data, dtype):
...
-        if archive:
-            self._save_to_archive(dtype_out_time)
-        logging.info('\t{}'.format(self.path_scratch[dtype_out_time]))
+        if save_files:
+            self._save_files(data, dtype_out_time)
+        if save_tar_files:
+            self._save_tar_files(dtype_out_time)
+        logging.info('\t{}'.format(self.path_out[dtype_out_time]))
 
     def _load_from_scratch(self, dtype_out_time, dtype_out_vert=False,
```
spencerahill: Let's rename this too... `_load_from_disk`?  All "scratch" references should be gone.

aospy/calc.py
```diff
@@ -897,7 +897,7 @@ def _load_from_scratch(self, dtype_out_time, dtype_out_vert=False,
 
     def _load_from_archive(self, dtype_out_time, dtype_out_vert=False):
```
spencerahill: Let's rename `_load_from_tar`.  Is any of this logic unique to /archive other than the `dmget` call?  For the time being, add a try/except guard around the dmget call here as well, in case anybody tries using this from outside GFDL.

spencerkclark:
> Probably at some point I screwed up the division logic, saw the tests were failing and slapped it on, and then forgot to change the tests whenever it was fixed again!

There must be a python version dependence here though, since it fails in python 2.7, but not python 3.5?  Maybe just skip them and fix things in a later PR?

spencerahill:
> There must be a python version dependence here though, since it fails in python 2.7, but not python 3.5? Maybe just skip them and fix things in a later PR?

Sorry, yet more rookie environment mistakes!  I didn't have pytest installed for 2.7, and didn't realize that even though I had `source activate`d my py27 env, it was still using the 3.5 pytest.  You're right, they fail on 2.7.  This is probably straightforward but not important right now (xref #50)...just do whatever you think is best.

One thing to keep in mind is that separate `runs.py`, `models.py`, etc. modules aren't strictly necessary: it could all be a single module, which may make more sense for small snippet-like projects such as the test one.

spencerahill:
Final comment, then let's merge: directory naming; let's rename `local_test_objs` to `data`, with subdirectories `netcdf` and `objects`.  (Just trying to emulate xarray's structure)

Another nice piece of work @spencerkclark, despite my best efforts to confuse you!  Thanks, as always I really appreciate it.

spencerkclark:
> One thing to keep in mind is that separate runs.py, models.py, etc. modules aren't strictly necessary: it could all be a single module, which may make more sense for small snippet-like projects such as the test one.
> 
> Final comment, then let's merge: directory naming; let's rename local_test_objs to data, with subdirectories netcdf and objects. (Just trying to emulate xarray's structure)

Good thoughts, I have modified things accordingly.  Perhaps the edits I needed to make to the path names to allow for this directory structure are not the most idiomatic, but they get the job done.

@spencerahill, if you're cool with those, I think we're all set!

spencerahill:
Awesome.  Sorry, one last nit: let's make the example object names more explicit: `example_proj`, `example_model`, `example_run`.  `im` is a little cryptic.  Sorry I didn't catch that before.

spencerahill:
Awesome, thanks @spencerkclark!

PR #106, opened 2016-11-02
Simple Jupyter Notebook Example

spencerkclark:
@spencerahill This is a very simple Jupyter notebook that outlines some example calculations using the sample data, following the naming conventions suggested in #1.  Let me know if you think I should add more descriptive comments to the notebook.  

You may view the notebook [here](https://github.com/spencerkclark/aospy/blob/templates/examples/example.ipynb).


spencerahill:
Perfect, thanks @spencerkclark!

PR #113, opened 2017-01-11
ENH improve logic & tests of limited date range workarounds

spencerahill:
This cleans up/refactors and slightly improves the logic for the limited daterange workaround.  (It was motivated by what I thought was a bug, but was actually just due to a filename of some netCDF data from the CAM model being misleading w/r/t its time values.)

@spencerkclark, one thing that's not strictly necessary that I changed here is using `open_mfdataset` rather than `open_dataset` followed by `xr.concat`.  Obviously the code is cleaner with mfdataset, but as I recall there were some performance problems there?  I can't find our discussion on it though either in the repo or our emails.  Is this still a problem?  I'm willing to revert to open_dataset if the performance hit is too much.


aospy/utils/times.py
```diff
@@ -168,6 +168,39 @@ def numpy_datetime_range_workaround(date):
...
+    -------
+    xarray.Dataset
+
+    """
+    time = ds[TIME_STR]
+    units = time.attrs['units']
+    units_yr = units.split(' since ')[1].split('-')[0]
+    min_yr = xr.decode_cf(time.to_dataset('dummy'))['time'].values[0].year
+    new_units_yr = pd.Timestamp.min.year + 2 - min_yr
+    new_units = units.replace(units_yr, str(new_units_yr))
```
spencerkclark, line 198: Very nice!  This is a clever way to avoid hard-coding January 1st as the start date.

aospy/utils/times.py
```diff
@@ -168,6 +168,39 @@ def numpy_datetime_range_workaround(date):
...
+    time.attrs['units'] = new_units
+    time.attrs['calendar'] = 'noleap'
```
spencerkclark: What is the reason for hard-coding a ```'noleap'``` calendar attribute here?

aospy/calc.py
```diff
@@ -459,33 +459,14 @@ def _create_input_data_obj(self, var, start_date=False,
...
-            # 'climatology_bounds' causes the ValueError b/c 'bnds' removed.
-            if self.dtype_in_time == 'av':
-                try:
-                    test = test.drop('climatology_bounds')
-                except ValueError:
-                    pass
-            test = xr.decode_cf(test)
-            ds_chunks.append(test)
-        ds = xr.concat(ds_chunks, dim=TIME_STR, data_vars='minimal')
+        vars_to_drop = ['time_bounds', 'nv', 'bnds', 'average_T1',
```
spencerkclark: I think we needed this due to an upstream bug in xarray (pydata/xarray#665), which has since been fixed.  Therefore I think we can remove this ```drop_variables``` part of the load step.

spencerkclark (commented):
> I changed here is using open_mfdataset rather than open_dataset followed by xr.concat. Obviously the code is cleaner with mfdataset, but as I recall there were some performance problems there? I can't find our discussion on it though either in the repo or our emails. Is this still a problem? I'm willing to revert to open_dataset if the performance hit is too much.

👍 I moved in that direction in #90.  To try and avoid the the performance hit due to inadvertent poor-chunking, I immediately do a ```arr.load()``` to load the data into memory before doing any computations -- so what I might do in this case is in the return statement (where the line is currently ```return arr```) I would make things ```return arr.load()```.

There is some overlap here with what's done in #90 but I'm happy to merge this as an intermediate step.  I particularly like your version of ```numpy_datetime_workaround_encode_cf```, and would like to incorporate it into #90.

aospy/calc.py
spencerahill: Good call!  Fixed.

aospy/utils/times.py
spencerahill: I was wondering the same thing!  It was in there previously, so I thought you had put it there for a particular reason.  Let's omit; don't see good reason why we would need noleap.

aospy/utils/times.py
spencerahill: Thanks!  It's predicated on the year and month being separated by a '-'.  I've never seen time units in a netCDF file formatted in any other way than that, but should just be aware of that choice moving forward.

spencerahill:
@spencerkclark thanks!  All good suggestions.  Pushing those edits now; will merge after.

Sorry for the duplication of your efforts!  This whole PR was inadvertent, based on me trying to fix what ended up not being a bug, and I lost track of what you were doing.

Issue #122, opened 2017-01-20
PNG image generated in the docs isn't working on RTD


spencerahill:
C.f. #121 and conversation therein

spencerahill:
Closed by #123 
Thanks @spencerkclark, nice find on xarray's fix!

PR #123, opened 2017-01-21
Try xarray fix for issue #122

spencerkclark:
I'm able to reproduce the "double-build" issue prior to this, and adding these couple lines to ```conf.py``` seems to fix things (it succeeds in finding the image after the first build).  Apparently it has been fixed upstream in IPython, but has just yet to be released.  

For reference here's the discussion in an xarray PR: pydata/xarray#1220
and the discussion in IPython is in: ipython/ipython#8733

@spencerahill want to try this?


spencerahill:
Thanks!  Yes, let's see how this does

PR #125, opened 2017-01-23
Major overhaul of our docs

spencerahill:
Definitely the coolest part is a logo!  Thanks to a friend of a friend with the requisite skills based on a rough design idea I came up with.  It can be tweaked in the future.

Despite our previous decision to focus less on the API docs and more on the top-level/narrative docs, the commits so far are largely the former.  I am still working on the latter and will add those commits when they're done.

@spencerkclark let me know any suggested edits.  I've tweaked RTD so that it builds from this branch, so we can see how things look in the wild.  Thanks!


aospy/model.py
```diff
@@ -16,16 +15,181 @@
...
+        description : str, optional
+            A description of the model.  This is not used internally by
+            aospy; it is solely for the user's information.
+        proj : {None, aospy.Proj}, optional
+            The parent Proj object.  When the parent `Proj` object is
+            instantiated with this Model included in its `models` attribute,
+            this will be over-written with that `Proj` object.
+        data_loader : aospy.DataLoader
+            The `DataLoader` object used to find the data on disk to be used
+            as inputs for aospy acalculations for this model.
```
spencerkclark: Typo: "aospy acalculations"

docs/data-structures.rst
```diff
@@ -0,0 +1,24 @@
+Data Structures
+===============
+aospy's data model comprises three core trunks.
+
+Core hierarchy
+--------------
+aospy's core data model is a hierarchy of classes, `Proj`, `Model`, and `Run`.
```
spencerkclark: No worries if you don't have time before tomorrow, but it might be nice at some point to explain a little more *why* one might want to use the boilerplate-heavy Proj-Model-Run framework.  In other words, beyond a more organized hierarchical directory structure for saved data, what does defining all three levels of objects get the user that they would not get otherwise?

I guess my rationale here is that there is the note that describes why the boilerplate heavy framework might be undesirable, but perhaps we should do a better job explaining the use case for when it is?

docs/data-structures.rst
```diff
@@ -0,0 +1,24 @@
...
+2. `Model`: This represents a single climate model, other numerical model, observational data source, etc.
+
+3. `Run`: This represents a single simulation, version of observational data, etc.
+
+.. note::
+   Currently, the Proj-Model-Run hierarchy is rigid, in that each Run has a parent Model, and each Model has a parent Proj.  For small projects, this can lead to a lot of boilerplate code.  Work is ongoing to relax this constraint to facilitate easier exploratory analysis.   
+
+Physical variables
+------------------
+The `Var` class is used to represent physical variables, e.g. precipitation rate or potential temperature.  This includes both variables which are directly available as netCDF files (e.g. they were directly outputted by your climate model) as well as those fields that must be computed from other variables (e.g. they weren't directly outputted but can be computed from other variables that were outputted).
```
spencerkclark: Minor point -- I might phrase things as "This includes both variables which are directly available *within* netCDF files..." rather than "as netCDF files," just to be clear that your variables don't need to be in separate files to be used in aospy.

docs/data-structures.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Specifying and executing calculations
+-------------------------------------
+The `Calc` class performs the calculations that a user specifies.  Unlike `Proj`, `Model`, `Run`, and `Var`, these objects are not intended to be serialized (i.e. saved on disk) for continual re-use.  Instead, they are generated as needed, perform their desired tasks, and then go away.
```
spencerkclark: Another minor point -- I see what you're trying to say here, but In the strict sense of the word, at least within the way we both currently use aospy, I would hesitate to say we "serialize" (e.g. via pickle) our Proj, Model, Run, and Var objects (though I suppose that would be possible).  Without going into too much more detail about "object libraries," maybe phrase things a little more generally as: "...these objects are not intended to be saved and packaged for continual re-use."

aospy/region.py
```diff
@@ -1,14 +1,145 @@
...
+             - True: apply the data's full land mask
+             - False: apply no mask
+             - 'ocean': mask out land rather than ocean
+             - 'strict_land': mask out all points that are not 100% land
+             - 'strict_ocean': mask out all points that are not 100% ocean
+
+        Examples
+        --------
+        Define a region spanning the entire globe
+        >>> globe = Region(name='globe', lat_bounds=(-90, 90),
```
spencerkclark, line 133: These examples aren't rendering properly in the built docs; I think you might just need a blank line separating the >>> block from the text.

aospy/region.py
```diff
@@ -1,14 +1,145 @@
...
+        ...                lon_bounds=(0, 360), do_land_mask=False)
+
+        Define a region corresponding to the Sahel region of Africa.  Because
+        this region crosses the 0 degrees longitude point, it has to be defined
+        using `mask_bounds` as the union of two lat-lon rectangles.
+        >>> sahel = Region(name='sahel', do_land_mask=True,
```
spencerkclark, line 142: Same as above

aospy/model.py
```diff
@@ -16,16 +15,181 @@
...
+            The model's name.  This must be unique from that of any other
+            `Model` objects being used by the parent `Proj`.
+        description : str, optional
+            A description of the model.  This is not used internally by
+            aospy; it is solely for the user's information.
+        proj : {None, aospy.Proj}, optional
+            The parent Proj object.  When the parent `Proj` object is
+            instantiated with this Model included in its `models` attribute,
+            this will be over-written with that `Proj` object.
+        data_loader : aospy.DataLoader
```
spencerkclark: Mark this as an "optional" argument for a Model; in fact, I would argue that at least the way things are structured right now, for a user I would recommend not bothering specifying a DataLoader to a Model object, because at best it would only prevent you from needing to specify a DataLoader for a single Run (and in that case you might as well just specify that at the Run level).

My fault for not being clearer about that earlier.

docs/api.rst
```diff
@@ -1,126 +1,197 @@
...
+.. autoclass:: aospy.region.Region
     :members:
     :undoc-members:
-    :show-inheritance:
 
-aospy.run module
-----------------
+    .. automethod:: aospy.region.Region.__init__
+
+Calculations
```
spencerkclark, line 119: If you have time, could you also flesh out the ```CalcInterface``` doc-string?  Currently one needs to be familiar with those arguments to create ```Calc``` objects.

spencerkclark (commented):
@spencerahill this looks really nice!  Thank you for all your work here (not just on the docs but also refactoring the static methods).  I have a few minor comments / suggestions.

The logo really does look nice, and gives the docs home page a much more polished feel!

spencerahill:
@spencerkclark thanks a lot for this review; will address it all.  And, like I said, I have more that I have yet to commit (still drafting/revising).

Nevertheless, I must confess that I misjudged both my time available and how long this would take, such that the docs won't be as polished as I hoped by the time I release v0.1 tonight.  But we can continue updating them and RTD will update accordingly without waiting for e.g. a v0.1.1 release.

I should have more time starting Wednesday to work on them further.  Also, anybody who wants to use aospy based on my talk will probably not be able to until after they get home from the meeting, giving us an additional buffer.  I'm also intending on announcing v0.1 within a week on the pangeo and xarray mailing lists.

aospy/model.py
spencerahill: Thanks for clarifying.

What about as a "template" that the data loaders of the children Run objects inherit from?  E.g. if all simulations are the same duration, specify the Model dataloader and then pass it to all the children (although I guess that could just as well be done with the dataloader of one of the Runs)

So what's the point of having a model-level (or proj-level) DataLoader at all?

docs/data-structures.rst
spencerahill: Yes, this is forthcoming.

aospy/model.py
spencerkclark: I *think* the way things are set up now prevents this ([see this comment in #97](https://github.com/spencerahill/aospy/pull/97#issuecomment-263038018)):

The issue currently has to do with a circular dependency (the constructor of the model object depends on Run objects already being instantiated), so one cannot reference that template (defined in the Model), when constructing the Run.  In order to leverage a "template" DataLoader right now, one would have to first create the template DataLoader where the Runs are created, and then use the alternative constructor for a GFDLDataLoader (the only one that supports this currently) that takes a template DataLoader as its first argument in each of the Runs that "inherit" it.

> So what's the point of having a model-level (or proj-level) DataLoader at all?

In hindsight, yes, it doesn't seem to make much sense to allow a Model level specification of a DataLoader (at least for how the code operates now).  You are welcome to remove it completely from the Model object as part of this PR if you'd like.

spencerkclark:
> Nevertheless, I must confess that I misjudged both my time available and how long this would take, such that the docs won't be as polished as I hoped by the time I release v0.1 tonight. But we can continue updating them and RTD will update accordingly without waiting for e.g. a v0.1.1 release.

Not a problem.  This is a lot of work, and I agree, there is no need to rush.

aospy/model.py
spencerahill: Ok I went ahead and did this.  Thanks for the explanation!

spencerkclark: Thanks for doing that; my apologies for setting things up that way to begin with.

spencerkclark:
@spencerahill the updates you've made so far look great.  If you'd like to merge things before tomorrow, that's cool with me (you can always pick things up with another PR later).  Though I suppose since these changes are on the live docs already it won't really make a difference, so it's completely up to you!

spencerahill:
@spencerkclark thanks, I'll go ahead and merge this.

I decided to include the main script that I use, as I realized it doesn't really make sense to document/explain the package without that core functionality.

The script itself definitely could be cleaned up -- in particular, the classes should be moved to a module within aospy proper (in addition to potentially being revamped).

It has gotten late, so it's possible I've made some careless mistakes...please ping if you catch any.  But I'm going ahead with the v0.1. release in the meantime; any fixes will go into v0.1.1.

PR #126, opened 2017-01-26
BUG Fix leftover references to ETA_STR

spencerkclark:
Some references to ```ETA_STR``` were unintentionally not converted to ```internal_names.ETA_STR``` as part of #90.  This short PR fixes this bug.


spencerahill:
Good catch.  Thanks!

spencerahill:
This is small enough that we can skip it, but let's remember moving forward to add What's New entries for all new bug fixes and features.

spencerkclark:
@spencerahill a quick and dirty unit test for this might be to include some three-dimensional test data in our ```test/data/netcdf/*.nc``` files, and compute a vertically integrated quantity (say column integrated water vapor for example) in ```test_calc_basic.py```.  It's not obvious to me how to pin this down to a test of a single method in ```calc.py```.

spencerkclark:
I fully agree that we need to have at least *some* Travis coverage here, and not rely on us simply running things ourselves every time we update the code.  @spencerahill what are your thoughts?  Does my suggestion above seem reasonable?

spencerahill:
@spencerkclark that seems like a good approach, although make sure that it's only a single timestep for package size reasons.

>It's not obvious to me how to pin this down to a test of a single method in calc.py.

That's what I was thinking also.

spencerahill:
I opened #129 to track this further

PR #133, opened 2017-01-31
Fix GitHub links in docs

spencerkclark:
It turns out that we need to configure the ability to link to issues and pull requests in our ```conf.py``` file.  I think this should do the trick (taken from xarray).


spencerkclark:
Sorry, my bad, I forgot a comma...

spencerahill:
Ha yes, just got the RTD build fail emails.  Will you push a new commit?

Issue #132, opened 2017-01-30
BUG: Problems handling data consisting of a single time slice


spencerkclark:
There are numerous problems caused (in many parts of the code) if we load in data that has only one time value; I've run into this in trying to construct a test case for #129 with a file consisting of a single monthly mean. 

I *think* it mostly has to do with the fact that in that case time is treated as a scalar coordinate and not an indexable dimension.  Since there are numerous places where we try to do ```da.sel ```or ```da.isel``` on the time dimension throughout the code, we run into a lot of problems.

@spencerahill is there a way to enforce that a coordinate is always indexable (in other words make sure that it is always in ```ds.dims``` even if it only has one value)?

spencerahill:
What immediately comes to mind for me is for us to just check upon loading if there is only a single time value.  

I.e. so `data_loader._load_data_from_disk` would become:

```python
    ...
    apply_preload_user_commands(file_set)
    ds =xr.open_mfdataset(file_set, preprocess=rename_grid_attrs,
                             concat_dim=internal_names.TIME_STR,
                             decode_cf=False)
    return _ensure_indexable_time(ds)

def  _ensure_indexable_time(ds):
    """"Force dataset to have indexable time coordinate.

    The primary use-case is for datasets with a single time-value, which
    results in a scalar coordinate that is therefore not indexable.
    """
    if <insert needed bool here>:
        return <insert needed logic here>
    else:
        return ds
```

Do you think something like that would work?

spencerahill:
Unless this becomes a major can of worms, let's get this into the v0.1.1 bugfix release; it's a not-too-uncommon use-case to have a single time value.

spencerahill:
Closed by #135

Issue #127, opened 2017-01-26
v0.1.2 announcement emails


spencerahill:
Once we've got v0.1 available via conda-forge (see #112), we should make an announcement in as many places as we can.  Below are the ones that come to mind for me.  @spencerkclark any others?

- [x] pangeo mailing list
- [x] xarray mailing list
- [x] pyaos mailing list
- [x] Princteon AOS mailing lists
- [x] GFDL python mailing list
- [x] My twitter (@spencerahill)
- [x] climlist

spencerkclark:
Does v0.1 ship with the code pre-#126?  As small as the changes are, without them the code is pretty broken for any calculations involving pressure (symptomatic of our lack of test coverage there...).  

I'm in the process of updating some of my object library and doing some calculations with the latest version (that's how I noticed the issues fixed in #126).  Do you think it might be worth holding off a few more days in case more bugs like this surface (and releasing an intermediate version with the bug fixes)?

spencerkclark:
I'm sorry about this; the issues in #126 stem completely from my refactoring of ```__config__``` and ```internal_names``` in #90.

spencerahill:
>Does v0.1 ship with the code pre-#126

> Do you think it might be worth holding off a few more days in case more bugs like this surface (and releasing an intermediate version with the bug fixes)?

Yes and yes.  So this will go into v0.1.1.  It has to be a formal PyPI and Github release to go on conda-forge.

No worries!  It happens, and that's on us collectively since we don't have sufficient tests in place to catch it.  Plus it was my suggestion to do `internal_names`, AND I missed it when I did the code review of your PR.

How hard will it be to write a unit test for this?  Were we premature in closing #22?  I would really like to get a test in that catches this.  We'll also need to add a what's new entry.

spencerkclark:
@spencerahill once we merge #135, give me another chance to run some real calculations (maybe it would be good if you did as well) just to be absolutely sure there are no more glaring bugs.  Then we can go ahead with the announcement emails.

spencerahill:
Yes, will do also.  Let's play with it through the end of this week; if no further critical problems are found then I'll submit to PyPI and conda-forge over the weekend or next week.  Will wait to do the announcements until we're uploaded to conda-forge.

spencerahill:
I'll also take care of #124 before then; that's the last remaining blocker for v0.1.1

spencerahill:
[...]

PR #135, opened 2017-01-31
Enable use of data of a single time slice

### LINKED ISSUE #129: Add tests to catch bug fixed by #126
See discussion in #126

spencerkclark:
Closes #129, #132

@spencerahill this is working (at least for the test suite, which now includes a case with a single time slice); however I'd like to at the minimum add a test for ```utils.times.ensure_time_as_dim```.  This is loosely what you had sketched out in #132; I had to make some changes in ```model.py``` as well.  Let me know what you think of this approach.

Remaining tasks:
- [x] Test of ```utils.times.ensure_time_as_dim```
- [x] What's new entry


aospy/utils/times.py
```diff
@@ -497,3 +497,40 @@ def assert_matching_time_coord(arr1, arr2):
...
+    ds : Dataset
+        Dataset with a time coordinate
+
+    Returns
+    -------
+    Dataset
+    """
+    TIME_STR = internal_names.TIME_STR
+    GRID_ATTRS_NO_TIMES = internal_names.GRID_ATTRS.copy()
+    del GRID_ATTRS_NO_TIMES[TIME_STR]
```
spencerahill: I was going to say that the `pop` method is more idiomatic here, but this SO question seems to imply otherwise: http://stackoverflow.com/questions/27155819/delete-a-key-and-value-from-an-ordereddict

Not a big deal either way

aospy/model.py
```diff
@@ -235,21 +235,25 @@ def _set_mult_grid_attr(self):
                 grid_attr = _get_grid_attr(grid_objs, name)
                 if grid_attr is not None:
                     # Rename coordinates to aospy's internal names.
+                    TIME_STR = internal_names.TIME_STR
                     renamed_attr = _rename_coords(grid_attr)
+                    if ((TIME_STR not in renamed_attr.dims) and (TIME_STR in renamed_attr)):
+                        renamed_attr = renamed_attr.drop(TIME_STR)
                     # Hack: MERRA data ending up with time-defined lat- and
```
spencerahill: Actually, can we get rid of this MERRA-related logic entirely?  If/when it comes up again, we can do more cleanly.

spencerahill:
I agree, definitely need unit tests for the new functionality.  Otherwise this looks solid.  I've added a few misc. in-line comments also.

aospy/utils/times.py
spencerkclark: Maybe we should move this logic to ```internal_names```?  It doesn't really make sense to construct this ```OrderedDict``` whenever this method is called.

spencerahill: Good call, let's do that.

aospy/utils/times.py
```diff
@@ -497,3 +497,40 @@ def assert_matching_time_coord(arr1, arr2):
...
+    del GRID_ATTRS_NO_TIMES[internal_names.TIME_BOUNDS_STR]
+    del GRID_ATTRS_NO_TIMES[internal_names.TIME_WEIGHTS_STR]
+
+    if TIME_STR not in ds.dims:
+        _time = [ds[TIME_STR].values.item()]
```
spencerkclark: Also should we make a general utility method to convert a scalar coordinate into a dimensional one?  I don't know if we'll ever use it anywhere else, but it might make this method a little more readable.
```python

def convert_scalar_coord_to_dim(ds, scalar_coord):
    """Create a DataArray with the given scalar coordinate as a dimension

    Parameters
    ----------
    ds : Dataset
        Dataset containing the scalar coordinate
    scalar_coord : str
        Name of the scalar coordinate

    Returns
    -------
    DataArray
    """
    data = [ds[scalar_coord].values.item()]
    da = xr.DataArray(data, coords=[data], dims=[scalar_coord])
    da.attrs = ds[scalar_coord].attrs
    return da
```

spencerahill: Yes, that's great.  The name is a little confusing though -- it reads as if you are getting rid of a coordinate and turning it into a dimension.

Also add a sentence or two similar to the time-specific docstring explaining why this is necessary, i.e. it's needed for indexing.

spencerkclark: Sure thing, I gave it another try in 23b7f27; let me know if that's any better

spencerkclark:
@spencerahill all set for another review!

spencerahill:
Great, thanks.  FYI tied up for a couple hours; might not get back to this and your suggestion on #127 until tonight.

spencerkclark:
@spencerahill rather than create a new branch, I've also added a couple cleanup commits to the example main script and ```CalcInterface``` docstring to highlight the new ```time_offset``` option.

No rush on reviewing this by the way!

spencerahill:
@spencerkclark this is great, merging now.

Thanks for the fast turnaround on all of this.

PR #141, opened 2017-02-03
TEST Add python 3.4 and 3.6 to Travis CI

### LINKED ISSUE #124: Make python versions supported (2.7, 3.4, 3.5, and 3.6) consistent across CI, setup.py, and conda releases
Should be straightforward, c.f. https://github.com/pydata/xarray/pull/1183

spencerahill:
Partially fixes #124; will address other parts in subsequent commits


spencerahill:
Travis good, now testing Appveyor

spencerahill:
Travis and Appveyor now good on 2.7 and 3.4-3.6.  
Also updating readthedocs; that build is running now.

conda-forge fix will have to be done within the staged-recipes/aospy repo

spencerahill:
RTD doc builds are good too.  So all set.
@spencerkclark FYI adding all three python 3 variants now makes the total CI time ~10 minutes.  I think most of the slowdown is w/ AppVeyor; it does each python variant one after the other.

Let's leave it as is for now, but if this slowdown becomes a pain we can consider switching back to a single 2.7 and 3.x test on appveyor.  Especially since we don't expect a huge fraction of users being on Windows.

PR #146, opened 2017-02-16
Update AppVeyor YML to use Travis conda environment files

spencerkclark:
Per https://github.com/spencerahill/aospy/issues/124#issuecomment-279788983

This may take a few tries to get right as I'm not familiar with the idiosyncrasies of shells on Windows (apologies in advance), but hopefully it won't be too hard.


spencerkclark:
@spencerahill well I'll be...it seems like I got this working on the first push.  Let me know if this looks good to you!

spencerahill:
Love when that happens 😄 

Thanks!

PR #153, opened 2017-03-14
DOCS conda preferred method except on python3.4

### LINKED ISSUE #147: Update install docs: conda-forge doesn't support 3.4
C.f. https://github.com/conda-forge/aospy-feedstock/pull/3 and https://github.com/conda-forge/aospy-feedstock/pull/2

In order to reduce their build load, conda-forge now only builds for 2.7, 3.5, and 3.6.  So we have to note in the docs that to install on 3.4 requires using pip rather than conda.

spencerahill:
Closes #147


spencerkclark:
Looks great!  

👍 for linking to Ryan's blog post in the docs;  installing one's own python environment is obviously a necessity to use ```aospy``` (or any other new python package for that matter), and I think Ryan did a great job describing how to do that in his post.

Issue #152, opened 2017-03-08
Move all except required user input out of main into aospy-level module


spencerahill:
The main script currently has multiple classes, e.g. `MainParams`, that don't belong there.  Since the script is the primary way we want users to interface with aospy, it should be stripped of everything except the options that the users should modify.

In so doing, we should also implement at least some basic tests of this functionality, and if possible clean up some of the logic, which is definitely not my best work.  But the priority now, given the desire to release soon, is more on cleaning up the front-end than the back-end, so long as the script works as we intend it to.

spencerahill:
Closed by #155

Issue #150, opened 2017-03-08
Update Examples page of docs: use main script to submit calculations


spencerahill:
The [Examples page](http://aospy.readthedocs.io/en/latest/examples.html#computing-the-annual-mean-total-precipitation-rate) of our docs uses `Calc` directly via `CalcInterface`, which is not our intended use-case for aospy.  We should modify this so that the examples/tutorials use the main script, since that *is* the intended use-case.

spencerahill:
Closed in #164

Issue #145, opened 2017-02-10
User config / set_options functionality is needed


spencerahill:
C.f. #69 and others.

Potential options
-----------------
- `set_options`, c.f. xarray and pandas; see https://github.com/spencerahill/aospy/issues/69#issuecomment-277345336
- config file, c.f. [matplotlibrc](http://matplotlib.org/users/customizing.html); i.e. users could have a `~/.aospyrc` file
- (I'm not sure these are mutually exclusive, i.e. options set interactively via set_options could take precedence over values in the config file)

Some places where this is needed
----------------------------------
- dmget/hsmget for data on GFDL's archive (and pre-load commands more generally)
- Desired backend for xarray.to_netcdf

Surely there are others also (please add them to this list).

How to proceed
----------------
We first need to decide how we want to support this.  Custom user configs are ubiquitous, and I don't see any reason why we would present any particularly tricky cases, so 

One potential snag is how to split user-specific things between aospy config and the user's own object library.  Maybe that could just go into the hierarchy: set_options > object library > .aospyrc.

This is all a little half baked...but this is an important piece of functionality moving forward and so I thought it should get it's own issue

spencerahill:
C.f. discussion in #156, considering adding a ObjectLibrary level above Proj in the hierarchy.

This could potentially be the place (and only place) where user library-wide settings are specified.  One benefit would be that this would enable using the same syntax as setting default values etc. at the lower levels in the hierarchy.

(Regardless of if we go this route, we should probably heed the Zen of Python and, at least to start, implement "one and only one" way of doing this.)

Issue #3, opened 2015-06-24
Var.vars should accept Vars that are themselves functions of multiple Vars


spencerahill:
Currently, if Var is passed a value for `func` when instantiated, it then populates the `vars` attribute with the Var objects that are passed into the given function.  But this is only supported for one level: the Var objects passed to the function must be stand-alone, i.e. don't require a function of their own.

The result is the unnecessary proliferation of functions.  For example, my flux divergence calculations currently have MSE, DSE, c_pT, gz, and general versions.  Each one takes as arguments all the arguments required for the pre-calculation (e.g. temp, hght, and sphum for MSE), computes it by a call to that function, then calls the general function with that output as the argument.

Possible solution: when a Calc object encounters a Var with nested functions such as this, it should, from bottom up, create Calc objects that returns the full timeseries of the specified function.  At the next level up, i.e. which is taking in this computed data as one of its Var objects, the code that loads data from netCDF will have to be bypassed, since you already have the timeseries in hand that you want.

spencerahill:
Data for any Var should be searched for in the following order:

1) As aospy-calculated `ts` data, saved to disk, matching the specified time/data type/etc.
2) As model-outputted data saved to disk.
3) If not 1) or 2), compute it by loading each of the Vars it depends on and executing the function.

This would lead to a recursive creation of Vars and their associated data until the data for the top-level Var is ultimately created.

spencerahill:
@spencerkclark's comment on #44 :

> An issue I've encountered, particularly for more involved calculations which involve many variables, is that the way the calculation needs to be carried out varies depending on the model used. For a concrete example take computing the net column heating due to radiative, sensible, and latent heat fluxes. In a comprehensive GCM (like AM2) there is a term due to reflected shortwave radiation (swup_toa); this does not exist in the gray-atmosphere idealized moist model. In addition there are differences in how to handle the evaporative heat flux from the surface in each Model. Currently I address this by creating two separate calc functions and two separate variables linked with each calc function. This is rather inefficient, and I end up with two differently named variables representing the same physical quantity.
> 
> Another somewhat related situation that I encounter is that in some cases a variable is already computed in the model, while in others a variable needs to be computed from other variables. This is especially tricky to deal with when that variable is involved in another function; again I need to create separate functions and variables for each Model.
> 
[...]

PR #155, opened 2017-03-16
Revamp main script functionality and associated examples

### LINKED ISSUE #151: Add example_obj_lib.py file using objects from tutorial.ipynb and use in main script
Essentially just copy and paste the object definitions in examples.ipynb into a separate .py file that is itself a functional aospy object library.  That way users have a direct example of a working object library file from the outset that they can copy elsewhere and manipulate, rather than having to cut and paste from a notebook.

Also, we should populate the main script with this object library, such that users can submit the main script and execute example calculations immediately upon installing aospy.

spencerahill:
Closes #151 and #152.

@spencerkclark I would really appreciate a review on this within the next few days if at all possible.  (Sorry for the time crunch and that I accidentally did this on the same branch as my last PR #153, so the diffs from that are also showing up below) . Thanks in advance!

This is a major overhaul of the functionality relating to the main script, and also includes one minor change to the `Calc` API.

- `Calc.region` had been a dict of `{region.name: region}` items.  But there is no need for this; we were never using the key, and we can always just grab the name from the object itself.  Doing so made the main script overhaul easier
- Enable multiple `projs` to be iterated over in the same multi-calc-generation call.
- Rename `main.py` to `aospy_main.py` and the `main()` function to `submit_mult_calcs()`
- Make the names used by `submit_mult_calcs` and thus the main script more descriptive
- Delete `find_obj.py` and most of logic from `main.py`; move what remained of `main.py` to new module `automate.py`.

I am not completely satisfied with my new approach in automate.py, which involves separating the specifications for Calc into "core" and "aux" categories.  This whole business would be trivial (just call `itertools.product` on everything), were it not for our support of `'default'` and `'all'` for some of the specs.

Basically I think this is well suited for a more OOP approach: a `CalcSpec` abstract base class, with one implementation of it for each spec that gets passed to Calc.  Each implementation would specify how it is to be handled, e.g. where to find its 'default' or 'all' values or what data type to be expected (e.g. `dtype_out_time` is a tuple that gets permuted over *within* a Calc, not across Calcs).  This would make it easier to, for example, support 'default' and 'all' values for every spec.

BUT we are a bit time crunched in terms of the pending blog post, and I've already ended up down the rabbit hole as it is.  So, barring fundamental objections to the current approach, I'm looking for critiques there, and we can come back to this bigger picture stuff later.  Sound good?

This bigger picture stuff notwithstanding, still some outstanding items:

- [x] Tests
- [x] Better docstrings in automate.py
- [x] what's new

Once this is locked in I will address #150, which I've already started on.


aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+    pass
+
+
+def _get_obj_default_or_all(name, parent, attr_name, defaults_attr_name):
+    if name == 'all':
+        return getattr(parent, attr_name).values()
+    elif name == 'default':
+        return getattr(parent, defaults_attr_name).values()
+    else:
+        return getattr(parent, attr_name)[name]
```
spencerkclark: Not necessarily opposed, just want to confirm:  does this mean that we need to switch from using lists to dictionaries when we specify the ```models``` attribute of a ```Proj``` object or the ```runs``` attribute of a ```Model``` object?

E.g. go from this:
```python
example_model = Model(
    name='example_model',
    grid_file_paths=('/path/to/files'),
    runs=[runs.control, runs.conv_off],
    default_runs=[]
)
```
to this:
```python
example_model = Model(
    name='example_model',
    grid_file_paths=('/path/to/files'),
    runs=dict(control=runs.control, conv_off=runs.conv_off),
    default_runs=dict()
)
```

I'm assuming this was a decision made to eliminate the need for the ```find_obj.py``` module?  (I'm all for that decision, by the way)

spencerkclark (commented):
> Calc.region had been a dict of {region.name: region} items. But there is no need for this; we were never using the key, and we can always just grab the name from the object itself. Doing so made the main script overhaul easier

> Delete find_obj.py

+1 for both of these

> we can come back to [...] bigger picture stuff later. Sound good?

Agreed; we could easily get caught in a lengthy discussion here, but for now let's hold off.  

Still working through the PR; I need to head out now, but will return to this later today and tomorrow.

aospy/automate.py
spencerahill: This is actually already taken care of via the [`utils.io.dict_name_keys`](https://github.com/spencerahill/aospy/blob/develop/aospy/model.py#L194) function.  So the list syntax still works.

spencerahill:
>Still working through the PR; I need to head out now, but will return to this later today and tomorrow.

Thanks a lot.  Sorry again for the large scope and putting us into a bit of a time crunch.

aospy/automate.py
spencerahill: On that point, the main script and example object library from these commits should work; try running the main script (from inside the example directory so that you don't have to install the object library).

spencerkclark: Ah I see this gets addressed automatically, should have noticed that sorry -- should we add some checking to ```utils.io.dict_name_keys``` to make sure that there are no duplicate names in the list of provided core objects (at least warn, maybe raise an exception)?  

For example, right now if you had two Runs with the same name attribute in the list provided, the first one would get overwritten by the second without the user knowing. 

> On that point, the main script and example object library from these commits should work; try running the main script (from inside the example directory so that you don't have to install the object library).

Indeed, I can confirm this works!

examples/example_obj_lib.py
```diff
@@ -0,0 +1,130 @@
...
+    'example_proj',
+    direc_out='example-output',
+    tar_direc_out='example-tar-output',
+    models=(example_model,),
+    regions=(globe, tropics)
+)
+
+
+projects = {example_proj.name: example_proj}
+variables = {var.name: var for var in [precip_largescale, precip_convective,
```
spencerkclark: This is somewhat big picture, but I'll just note here that this requirement (creating a dictionary mapping Var name attributes to Var objects) breaks my current object library, because I have some Var objects that have the same name attribute, but are different Var objects (to accommodate cases where a variable is native in one model, but needs to be computed from other variables for output from another model, see https://github.com/spencerahill/aospy/issues/3#issuecomment-158840815).

E.g. these two variables for OLR:
```python
olr = Var(
    name='olr',
    alt_names=('rlut',),
    units=units.W_m2,
    domain='atmos',
    description='All-sky outgoing longwave radiation at TOA.',
    def_time=True,
    def_vert=False,
    def_lat=True,
    def_lon=True
)

olr_imr = Var(
    name='olr',
    domain='atmos',
    description=('Outgoing longwave radiation'),
    variables=(swdn_sfc, vert_int_tdtsw_rad_imr, netrad_toa_imr),
    def_time=True,
    def_vert=False,
    def_lat=True,
    def_lon=True,
    func=calcs.idealized_moist_rad.energy.olr,
    units=units.W_m2
)
```
This will take some effort/thinking to address and I don't think it's necessarily a blocker here.  There's nothing that precludes me (or anyone else) from rolling their own solutions to a main script or object library (so until we have a better solution for #3 I may need to just continue using ```find_obj.py``` and my old main script to get things done).

spencerkclark: On a more general note, this practice of requiring the user to create module level dictionaries to set up their object libraries feels a little non-ideal;  again though, without more thought I don't have a good alternative solution.   For small object libraries it feels OK, but for larger ones (that get spread out across multiple submodules, like what we use in practice), these would need to be placed (somewhat hidden) in an ```aospy_user/__init__.py``` file.

For variables in particular, if one has a lot defined in a module (e.g. in ```aospy_user/variables/__init__.py```) is there a straightforward pythonic way to create this dictionary without much typing?

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+                    # Use Calc's attr names as keys.
+                    obj_trees.append(dict(proj=project, model=model, run=run))
+        return obj_trees
+
+    def _permute_aux_specs(self):
+        # Drop the "core" specifications, which are handled separately.
+        specs = self._specs_in.copy()
+        [specs.pop(core) for core in self._CORE_SPEC_NAMES]
+        # TODO: don't just pull from first project.
+        regions = _get_mult_objs_by_names(specs['regions'], self._projects[0],
```
spencerkclark: Would this prevent a user from using a region that was not explicitly listed as a Region in a particular Proj object?

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+        return _permute_aux_specs(specs)
+
+    def _combine_core_aux_specs(self):
+        return _combine_core_aux_specs(self._permute_core_specs(),
+                                       self._permute_aux_specs())
+
+    def create_calcs(self):
+        return _create_calcs(self._combine_core_aux_specs())
+
+    def _print_specs(self):
```
spencerkclark: I think you'll need to update this function to use the ```self._specs_in``` dictionary, rather than the (now removed) ```projects```, ```models```, ```runs``` etc. attributes in order for it to run (though I'm guessing you're aware of this, since you've commented it out below).

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+        else:
+            out.append(o)
+    return out
+
+
+def submit_mult_calcs(calc_suite_specs, obj_lib, parallelize=False,
+                      prompt_verify=False, verbose=True):
+    """Generate and execute all specified computations."""
+    calc_suite = CalcSuite(calc_suite_specs, obj_lib)
+    # calc_suite._print_specs()
```
spencerkclark: Uncomment this once ```_print_specs()``` is updated

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+
+        # Convert to attr names that Calc is expecting.
+        calc_aux_mapping = self._NAMES_SUITE_TO_CALC_MAPPING.copy()
+        [calc_aux_mapping.pop(core) for core in self._CORE_SPEC_NAMES]
+        for suite_name, calc_name in calc_aux_mapping.items():
+            specs[calc_name] = specs.pop(suite_name)
+        return _permute_aux_specs(specs)
+
+    def _combine_core_aux_specs(self):
+        return _combine_core_aux_specs(self._permute_core_specs(),
```
spencerkclark: It's kind of confusing to have two functions of the same name that are defined differently.  Is there a good reason to define the module-level version of ```_combine_core_aux_specs```?  Could you simply just implement what it does within this class-level function?

spencerkclark: Maybe at the module level create a function called ```_merge_dicts``` and then here you can write a one-liner using ```itertools.product```?  Would that work?

E.g. at the module level:
```python
def _merge_dicts(a, b):
    merged = a.copy()
    merged.update(b)
    return merged
```

and at the instance-level (here):
```python
def _combine_core_aux_specs(self):
    return [_merge_dicts(a, b) for a, b in itertools.product(
            self._permute_core_specs(), self._permute_aux_specs())]
```

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _get_mult_objs_by_names(names, parent, attr_name, defaults_attr_name):
+    if isinstance(names, str):
+        names = [names]
+    return [_get_obj_default_or_all(name, parent, attr_name,
+                                    defaults_attr_name) for name in names]
+
+
+def _permute_aux_specs(specs):
```
spencerkclark: I think you could write this function pretty compactly as a list-comprehension:

```python
def _permute_aux_specs(specs):
    return [dict(zip(specs.keys(), perm)) for perm in itertools.product(*specs.values())]
```

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+    all_specs = []
+    for core_dict in core_dicts:
+        for aux_dict in aux_dicts:
+            combined = core_dict.copy()
+            combined.update(aux_dict)
+            all_specs.append(combined)
+    return all_specs
+
+
+def _create_calcs(specs):
```
spencerkclark: I think this could also be written more compactly as a list-comprehension

aospy/automate.py
```diff
@@ -0,0 +1,220 @@
...
+        print(reset)
+
+
+def exec_calcs(calcs, parallelize=False):
+    if parallelize:
+        pool = multiprocess.Pool()
+        return pool.map(lambda calc: calc.compute(), calcs)
+    out = []
+    for calc in calcs:
+        try:
```
spencerkclark: Could there be a way to encapsulate this try-except logic such that it could also be used above when we parallelize the calculations?  That way it would ensure the same warning behavior through each pathway.

spencerkclark (commented):
@spencerahill I've finished making a first pass here -- overall I think it looks pretty good!  Thanks for taking this on.

A few high-level questions came up (related to #3), but we might want to put those off until later.

aospy/automate.py
spencerahill: >Indeed, I can confirm this works!

Great!

>For example, right now if you had two Runs with the same name attribute in the list provided, the first one would get overwritten by the second without the user knowing.

Good catch; will address.

>I think you could write this function pretty compactly as a list-comprehension

Yes, will do and the other list comprehensions you suggest

aospy/automate.py
spencerahill: Yes that's a good point...I need to think about this

A related issue (although I don't see an immediate use-case) is if the same name was used by Region objects in different projects but with different region definitions.  In that case this logic would incorrectly use the region definition for the first project in all cases.

spencerahill: >same name was used by Region objects in different projects but with different region definitions

This would remain a problem even if e.g. we formed the union of regions defined across all projects.  In that case the problem reduces to the same issue you raised above re: conflicting names of Var objects.

aospy/automate.py
spencerahill: Yes definitely

examples/example_obj_lib.py
spencerahill: >this requirement (creating a dictionary mapping Var name attributes to Var objects) breaks my current object library, because I have some Var objects that have the same name attribute, but are different Var objects

>this practice of requiring the user to create module level dictionaries to set up their object libraries feels a little non-ideal

These are both good points and I definitely want to avoid breaking your workflow.  Will have a separate comment and/or Issue on these...I have some ideas but need to clarify them.

spencerahill:
@spencerkclark thanks a lot for the review.  Your concerns largely coincide with the parts I was most uncomfortable about, so at least we're on the same page about what needs improving.

In terms of long-term solutions, I opened #156 to discuss the aspects re: how to register objects with minimal input by the users.

The other big issue you note that I still need to think about is what the appropriate "parent" is of Vars and Regions.  Will follow up with my thinking on that.

aospy/automate.py
spencerahill: I agree, I got carried away here with the module-level functions.  The original motivation was for ease of testing and potential re-use of functionality elsewhere, but this logic seems pretty specific to CalcSuite.

spencerahill:
@spencerkclark I *think* I have managed to relax the prior onerous requirement of defining {name: obj} dicts for the projects and variables.  CalcSuite now simply searches the object.

However, now I'm realizing this still doesn't work with either of our existing setups, because we are importing a `variables` module into the object library's top-level namespace.  In what I've just pushed, the `Var` objects are searched for at that top level.

Perhaps a good solution is a try/except case, wherein if 'variables' is defined it is used, and if not then this new method of finding all the `Vars` is used.  Does that make sense?

I added some initial tests; will add more.  C.f. our convo offline, you're welcome to add more, but more important for me is your feedback on the overall approach.

spencerahill:
>Perhaps a good solution is a try/except case, wherein if 'variables' is defined it is used, and if not then this new method of finding all the Vars is used

Edit: typo in the commit message; should be 'variables'

Just pushed this, let me know if it works for you.

spencerkclark:
> Perhaps a good solution is a try/except case, wherein if 'variables' is defined it is used, and if not then this new method of finding all the Vars is used. Does that make sense?

Thanks for the updates; yes, I think this is a very good compromise.  I will have a closer look at your changes tomorrow afternoon.

spencerahill:
Great, thanks!  I'll try to do some more testing in the meantime

spencerahill:
Just added some more tests.  This is my first time using pytest.  I immediately like it a lot, although I'm not certain I'm using it correctly.  I feel like maybe I'm overdoing it with all of the `@pytest.fixture`s, and also I haven't figured out the syntax of the one parameterized case.

@spencerkclark since I know you've been using pytest recently, any comments on these aspects would be appreciated.

spencerahill:
I have encountered another problem.  When I execute aospy_main the first time, it works fine.  But the second time, I'm getting an error related to the tar output:

```
~/Dropbox/py/aospy/examples$ ./aospy_main.py
INFO:root:Initializing Calc instance: Calc object: precip_total, example_proj, example_model, example_run (Wed Mar 22 10:11:03 2017)
INFO:root:Initializing Calc instance: Calc object: precip_convective, example_proj, example_model, example_run (Wed Mar 22 10:11:03 2017)
INFO:root:Initializing Calc instance: Calc object: precip_largescale, example_proj, example_model, example_run (Wed Mar 22 10:11:03 2017)
INFO:root:Initializing Calc instance: Calc object: precip_conv_frac, example_proj, example_model, example_run (Wed Mar 22 10:11:03 2017)
Calc object: precip_total, example_proj, example_model, example_run
Calc object: precip_convective, example_proj, example_model, example_run
Calc object: precip_largescale, example_proj, example_model, example_run
Calc object: precip_conv_frac, example_proj, example_model, example_run
Perform these computations? [y/n] y
INFO:root:Getting input data: Var instance "precip_largescale" (Wed Mar 22 10:11:06 2017)
/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/conventions.py:389: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  result = decode_cf_datetime(example_value, units, calendar)
/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/conventions.py:408: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  calendar=self.calendar)
INFO:root:Getting input data: Var instance "precip_convective" (Wed Mar 22 10:11:06 2017)
/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/conventions.py:389: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  result = decode_cf_datetime(example_value, units, calendar)
/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/conventions.py:408: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  calendar=self.calendar)
INFO:root:Computing timeseries for 0004-01-01 00:00:00 -- 0006-12-31 00:00:00.
INFO:root:Applying desired time-reduction methods. (Wed Mar 22 10:11:06 2017)
INFO:root:Writing desired gridded outputs to disk.
tar: Option --delete is not supported
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
INFO:root:	example-output/example_proj/example_model/example_run/precip_total/precip_total.ann.av.from_monthly_ts.example_model.example_run.0004-0006.nc
HDF5-DIAG: Error detected in HDF5 (1.8.17) thread 0:
  #000: H5F.c line 522 in H5Fcreate(): unable to create file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1024 in H5F_open(): unable to truncate a file which is already open
    major: File accessibilty
    minor: Unable to open file
  File "./aospy_main.py", line 110, in <module>
    prompt_verify=prompt_verify)
  File "/Users/shill/Dropbox/py/aospy/aospy/automate.py", line 247, in submit_mult_calcs
    return exec_calcs(calcs, parallelize=parallelize)
  File "/Users/shill/Dropbox/py/aospy/aospy/automate.py", line 230, in exec_calcs
    o = calc.compute()
  File "/Users/shill/Dropbox/py/aospy/aospy/calc.py", line 642, in compute
    save_files=save_files, save_tar_files=save_tar_files)
  File "/Users/shill/Dropbox/py/aospy/aospy/calc.py", line 707, in save
    self._save_files(data, dtype_out_time)
  File "/Users/shill/Dropbox/py/aospy/aospy/calc.py", line 663, in _save_files
    data_out.to_netcdf(path)
  File "/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/core/dataset.py", line 953, in to_netcdf
    unlimited_dims=unlimited_dims)
  File "/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/backends/api.py", line 563, in to_netcdf
    store = store_cls(path, mode, format, group, writer)
  File "/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/backends/netCDF4_.py", line 205, in __init__
    self.ds = opener()
  File "/Users/shill/Dropbox/miniconda3/lib/python3.5/site-packages/xarray/backends/netCDF4_.py", line 181, in _open_netcdf4_group
    ds = nc4.Dataset(filename, mode=mode, **kwargs)
  File "netCDF4/_netCDF4.pyx", line 1848, in netCDF4._netCDF4.Dataset.__init__ (netCDF4/_netCDF4.c:13992)
OSError: Permission denied
```

If I delete the tar output and re-submit, it works fine.  @spencerkclark does this happen for you?  It's possibly a MacOS thing, given the OSError and that we've not had this problem in the past with the tar output (I'm running from my Macbook Pro).

spencerkclark:
> Just added some more tests. This is my first time using pytest. I immediately like it a lot, although I'm not certain I'm using it correctly. I feel like maybe I'm overdoing it with all of the ```@pytest.fixtures```, and also I haven't figured out the syntax of the one parameterized case.

Thanks for adding some more tests!   I'm still getting the hang of pytest as well, but I also think it's pretty awesome.  I *think* your use of fixtures is actually what's causing your trouble with ```parametrize```.  As it turns out the results of fixture functions cannot be used within ```parametrize``` decorators; they can only be passed as arguments to test functions themselves.  So basically you'll just need to explicitly write out what ```obj_lib```, ```all_vars```, and ```all_projects``` should be there:
```python
@pytest.mark.parametrize(('type_', 'obj_lib', 'expected'),
                         [(Var, examples, {condensation_rain, convection_rain, precip, ps, sphum}),
                          (Proj, examples, {example_proj})])
def test_get_all_objs_of_type(type_, obj_lib, expected):
    actual = _get_all_objs_of_type(type_, obj_lib)
    assert expected == actual
```
and I think things should work. 

I'm taking a look at the rest of your changes now, and will see if I can fill in some gaps in the test coverage after.

aospy/automate.py
```diff
@@ -0,0 +1,247 @@
...
+                for run in runs:
+                    # Use Calc's attr names as keys.
+                    obj_trees.append(dict(proj=project, model=model, run=run))
+        return obj_trees
+
+    def _get_aux_specs(self):
+        # Drop the "core" specifications, which are handled separately.
+        specs = self._specs_in.copy()
+        [specs.pop(core) for core in self._CORE_SPEC_NAMES]
+        specs['regions'] = self._get_regions()
```
spencerkclark: I think this will *always* use every region defined by the user in their object library.

aospy/automate.py
```diff
@@ -0,0 +1,247 @@
...
+        specs['variables'] = self._get_variables()
```
spencerkclark: (Similar to above) I think this will *always* use every variable defined by the user in their object library.  Is that correct?

aospy/automate.py
```diff
@@ -0,0 +1,247 @@
...
+        projs = _get_all_objs_of_type(Proj, self._obj_lib)
+        names = self._specs_in['projects']
+        return set([proj for proj in projs if proj.name in names])
+
+    def _get_regions(self):
+        return [_get_all_objs_of_type(Region, self._obj_lib)]
+
+    def _get_variables(self):
+        if hasattr(self._obj_lib, 'variables'):
+            return self._obj_lib.variables
```
spencerkclark: I've gone ahead and tried this in my object library; this line does get called, but it causes a problem because this function needs to return an iterable to be used in [line 53](https://github.com/spencerahill/aospy/pull/155/files#diff-e3c273a553369da44ac80bfa80ca1eebR53). 

The obvious change (which I tried) would be to return the result of ```_get_all_objs_of_type(Var, self._obj_lib.variables)```, but this results in ```automate``` attempting to create calculations involving all the variables defined in my object library (ignoring the one I specified in the specifications), prompting my comments below.  Is there a simple way to address this?

spencerkclark:
> I have encountered another problem. When I execute aospy_main the first time, it works fine. But the second time, I'm getting an error related to the tar output:

I've actually had something similar happen (prior to this PR), and not for tar output but for standard netCDF output.   It only seems to occur for regional calculations.  Here's a minimal working example:
```python
In [1]: import xarray as xr

In [2]: ds = xr.Dataset()

In [3]: ds['a'] = 1

In [4]: ds['b'] = 2

In [5]: ds.to_netcdf('test.nc')

In [6]: ds.to_netcdf('test.nc')
``` 

For some reason I can call ```to_netcdf``` twice in the same session, but if I shut down ```ipython``` and restart the kernel, and reopen the file to try again, things fail:
```
In [1]: import xarray as xr

In [2]: ds = xr.open_dataset('test.nc')

In [3]: ds.to_netcdf('test.nc')
HDF5-DIAG: Error detected in HDF5 (1.8.17) thread 0:
  #000: H5F.c line 522 in H5Fcreate(): unable to create file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1024 in H5F_open(): unable to truncate a file which is already open
    major: File accessibilty
    minor: Unable to open file
OSError                                   Traceback (most recent call last)
<ipython-input-3-d7687fe1d488> in <module>()
----> 1 ds.to_netcdf('test.nc')

/home/skc/miniconda/envs/research/lib/python3.6/site-packages/xarray/core/dataset.py in to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims)
    951         return to_netcdf(self, path, mode, format=format, group=group,
    952                          engine=engine, encoding=encoding,
--> 953                          unlimited_dims=unlimited_dims)
    954
    955     def __unicode__(self):

/home/skc/miniconda/envs/research/lib/python3.6/site-packages/xarray/backends/api.py in to_netcdf(dataset, path, mode, format, group, engine, writer, encoding, unlimited_dims)
    561     sync = writer is None
    562
--> 563     store = store_cls(path, mode, format, group, writer)
    564
    565     if unlimited_dims is None:

/home/skc/miniconda/envs/research/lib/python3.6/site-packages/xarray/backends/netCDF4_.py in __init__(self, filename, mode, format, group, writer, clobber, diskless, persist)
    203                                    diskless=diskless, persist=persist,
    204                                    format=format)
--> 205         self.ds = opener()
    206         self.format = format
    207         self.is_remote = is_remote_uri(filename)

/home/skc/miniconda/envs/research/lib/python3.6/site-packages/xarray/backends/netCDF4_.py in _open_netcdf4_group(filename, mode, group, **kwargs)
    179     import netCDF4 as nc4
    180
--> 181     ds = nc4.Dataset(filename, mode=mode, **kwargs)
    182
    183     with close_on_error(ds):

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Dataset.__init__ (netCDF4/_netCDF4.c:13992)()

OSError: Permission denied
```
(This is on a GFDL workstation by the way, so not Mac OS specific)

spencerkclark:
I'm not sure if this issue in xarray is related: https://github.com/pydata/xarray/issues/1215

aospy/automate.py
spencerkclark: Also does this work if there is no region specified?  (In other words if someone doesn't want to do a spatial reduction on result?)

aospy/automate.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _get_projects(self):
+        projs = _get_all_objs_of_type(Proj, self._obj_lib)
+        names = self._specs_in['projects']
+        return set([proj for proj in projs if proj.name in names])
+
+    def _get_regions(self):
+        return [_get_all_objs_of_type(Region, self._obj_lib)]
+
+    def _get_variables(self):
```
spencerkclark: Would rewriting this function to look like this be acceptable?  This would basically revert the behavior back to using the python variable name for each Var (not its name attribute).  
```python
def _get_variables(self):
    objs = getattr(self._obj_lib, 'variables', self._obj_lib).__dict__
    variables = {
        name: var for name, var in objs.items() if isinstance(var, Var)}
    names = self._specs_in['variables']
    return set([var for name, var in variables.items() if name in names])
```
This is pretty confusing though, since the name attribute is used for every other core object type (though that's also the way things are handled in the main script prior to this PR).  

To be honest I think for my particular situation / use-case, writing a similar, simpler, kind of main script to handle permutation that asked the user to specify lists of actual aospy core objects (rather than strings) would be the way to go.  It's up to you what you want to do here; I'm not suggesting we switch to that in the current PR.  I'm more considering in the medium-term writing that type of main script on my own, and potentially thinking about adding it to the aospy core later.

spencerahill:
@spencerkclark big thanks for this batch of comments.  Sorry wasn't able to get to them tonight, will tomorrow for sure.  Once again you've caught some important things.

aospy/automate.py
spencerahill: Wow, both very good points, and same with `_get_variables` below.

spencerahill:
(Github not letting respond to this in-line for some reason)

>results of fixture functions cannot be used within parametrize decorators; they can only be passed as arguments to test functions themselves. So basically you'll just need to explicitly write out what obj_lib, all_vars, and all_projects should be there

Darn.  I kind of assumed that was the case but was hoping there was some trickery to avoid it.  Oh well.  Thanks for the commit fixing it.

aospy/automate.py
spencerahill: You're right, we should stop dinking around with this string silliness and use the actual objects.  I have started in on this.

spencerkclark: @spencerahill cool!  Please let me know if you need any help.

spencerahill:
@spencerkclark I've taken a stab at switching from names to objects and have managed to get back to roughly the same spot I was before otherwise.

I also added a "library" kwarg to the `calc_suite_specs` (essentially moving it there, rather than it being a separate kwarg of `submit_mult_calcs`).

Note that the `date_ranges` issue I note in the commit message goes away if you just make it `['default']`.  But with that fixed I'm noticing that I'm generating too many Calcs; it's probably iterating over the individual regions.  Should be easy enough to fix these.

FYI I'll be away for the next hour or so but this is my sole task otherwise tonight.  Will try my best to get all outstanding problems addressed before tomorrow.

aospy/automate.py
spencerahill: Good idea, fixing this makes that logic much cleaner

spencerahill:
>I've actually had something similar happen (prior to this PR), and not for tar output but for standard netCDF output.

Yikes.  Is this occurring when you use aospy?  Thanks to your MWE, it definitely does look like an upstream problem.  Nevertheless, we need to think of the best way to mitigate.  At the very least, I think we'd need one or more Warning boxes in the docs noting this issue and your workaround of staying within the same session.

(I haven't come across this myself, although I've been working largely outside of aospy lately.)

spencerahill:
Ok  @spencerkclark I believe I have addressed everything besides the upstream OSError issue.  Still need to flesh out the tests, and I won't merge until I do so.  But hopefully we're getting close here.

spencerkclark:
> Yikes. Is this occurring when you use aospy? Thanks to your MWE, it definitely does look like an upstream problem. Nevertheless, we need to think of the best way to mitigate. At the very least, I think we'd need one or more Warning boxes in the docs noting this issue and your workaround of staying within the same session.

Yes, I encountered this when using aospy.  Unfortunately staying in the same session is not really a viable workaround (this issue has mainly come back to bite me when I (intentionally or unintentionally) try to overwrite some regional calculations days or even weeks after I first created them).  Another workaround is to use the ```engine='scipy'``` keyword argument in the call to ```to_netcdf```, though I wouldn't necessarily use this as the default, because it would add an extra dependency.  We'd need to accelerate a solution for #145 to make setting this option user-friendly, however.

aospy/test/test_automate.py
```diff
@@ -0,0 +1,187 @@
...
+    return {globe, sahel}
+
+
+@pytest.fixture
+def calcsuite_specs():
+    return {
+        'time_offset': [None],
+        'date_range': ['default'],
+        'intvl_in': ['monthly'],
+        'region': [(globe, sahel)],
```
spencerkclark: For efficiency reasons it's good to execute all regional reductions within a single Calc object (so it makes sense under the hood to pass a sequence of Region objects to a Calc constructor for each variable), but from a user's perspective do you think this nested syntax could be confusing?  

For instance *I think*, given the way we are doing the iteration (i.e. full ```itertools.product```), under the current setup one could alternatively write this keyword argument as ```'region': [(globe,), (sahel,)]``` and it would produce the same result (it would just take roughly twice as long, because two separate Calc objects would be created for doing the two different regional reductions). 

Could we instead simplify things so that one just provides a simple one-dimensional list?

aospy/test/test_automate.py
```diff
@@ -0,0 +1,187 @@
...
+        'dtype_out_time': [('av', 'reg.av')],
```
spencerkclark: I think the same concern above applies here

aospy/automate.py
```diff
@@ -0,0 +1,263 @@
...
+                        self._NAMES_SUITE_TO_CALC[_PROJECTS_STR]: project,
+                        self._NAMES_SUITE_TO_CALC[_MODELS_STR]: model,
+                        self._NAMES_SUITE_TO_CALC[_RUNS_STR]: run,
+                    })
+        return obj_trees
+
+    def _get_regions(self):
+        """Get the requested regions."""
+        if self._specs_in[_REGIONS_STR] == 'all':
+            try:
```
spencerkclark: Instead of a try and except block, should we instead use ```getattr``` with a default argument?  

E.g. ```_get_all_objs_of_type(Region, getattr(self._obj_lib, 'regions', self._obj_lib))```

aospy/automate.py
```diff
@@ -0,0 +1,263 @@
...
+                return [_get_all_objs_of_type(Region, self._obj_lib.regions)]
+            except AttributeError:
+                return [_get_all_objs_of_type(Region, self._obj_lib)]
+        else:
+            return set(self._specs_in[_REGIONS_STR])
+
+    def _get_variables(self):
+        """Get the requested variables."""
+        if self._specs_in[_VARIABLES_STR] == 'all':
+            try:
```
spencerkclark: Same as above, should we use ```getattr``` here?

aospy/automate.py
```diff
@@ -0,0 +1,263 @@
...
+            specs[calc_name] = specs.pop(suite_name)
+        return _permuted_dicts_of_specs(specs)
+
+    def _combine_core_aux_specs(self):
+        """Combine permutations over core and auxilliary Calc specs."""
+        all_specs = []
+        for core_dict in self._permute_core_specs():
+            for aux_dict in self._permute_aux_specs():
+                all_specs.append(_merge_dicts(core_dict, aux_dict))
+                print(_merge_dicts(core_dict, aux_dict), '\n')
```
spencerkclark: For now, in lieu of a nicer repr, could we use [```pprint.pprint```](https://docs.python.org/3/library/pprint.html#pprint.pprint) from the standard library here (and elsewhere) for nicer printing of dicts?

E.g. ```pprint.pprint(_merge_dicts(core_dict, aux_dict))``` produces:
```
{'date_range': (datetime.datetime(3, 1, 1, 0, 0),
                datetime.datetime(6, 12, 31, 0, 0)),
 'dtype_in_time': 'ts',
 'dtype_in_vert': 'sigma',
 'dtype_out_time': ('av', 'reg.av'),
 'dtype_out_vert': None,
 'intvl_in': 'monthly',
 'intvl_out': 'ann',
 'model': Model instance "idealized_moist_rad",
 'proj': Project instance "itcz",
 'region': (Geographical region "globe",),
 'run': aospy.Run instance "control",
 'time_offset': {'days': -15},
 'var': Var instance "precip_im"}
```

spencerkclark (commented):
@spencerahill pending tests, this is looking very close!  I have one more high-level concern (but I don't think it would be too hard to address if you feel it's valid).

I will see if I can work on a few tests now.

spencerkclark:
Thanks for all of your work on this by the way!

aospy/automate.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def __init__(self, calc_suite_specs):
+        self._specs_in = calc_suite_specs
+        self._obj_lib = self._specs_in[_OBJ_LIB_STR]
+
+    def _get_requested_spec(self, obj, spec_name):
+        """Helper to translate user specifications to needed objects."""
+        requested = self._specs_in[spec_name]
+        if isinstance(requested, str):
+            return [_get_attr_by_tag(obj, requested, spec_name)]
```
spencerkclark: Since the aospy core variable attributes use ```utils.io.dict_name_keys```, this returns a list of a single dictionary, while the line below the ```else``` statement returns a list of aospy core objects (this inconsistency causes problems).  Should we do away with using ```dict_name_keys``` in the core objects themselves, and just assign lists to their ```models``` and ```runs``` attributes?

spencerkclark: Working on implementing this now; I'm happy to roll back if you're against this change.

aospy/test/test_automate.py
```diff
@@ -0,0 +1,192 @@
...
+        'dtype_out_vert': [None]
+    }
+
+
+@pytest.mark.parametrize(
+    ('obj', 'tag', 'attr_name', 'expected'),
+    [(example_proj, 'all', _MODELS_STR, [example_model]),
+     (example_proj, 'default', _MODELS_STR, [example_model]),
+     (example_model, 'all', _RUNS_STR, [example_run]),
+     (example_model, 'default', _RUNS_STR, [])])
```
spencerkclark: Right now by default if one does not specify a ```default_runs``` list in a Model constructor, an empty list is created for the ```default_runs``` attribute, but for a Proj object, if no ```default_models``` list is provided, the ```default_models``` attribute is set to all the Model objects.  I think we should try to be consistent; which behavior do you prefer?

aospy/automate.py
spencerahill: No you're right, this is the way to go.  Thanks for the fix.

aospy/test/test_automate.py
spencerahill: Yes, we should definitely be consistent.  I am leaning towards default returning empty if it hasn't been set.  That feels more intuitive.  But I could be convinced the other way, if you feel strongly about it.

aospy/automate.py
spencerahill: Yes, much cleaner

spencerahill:
>Another workaround is to use the engine='scipy' keyword argument in the call to to_netcdf, though I wouldn't necessarily use this as the default, because it would add an extra dependency. 

If using the scipy engine solves this, I definitely think it's worth the dependency.  Our preferred installation method of conda already installs it anyway due to (presumably) xarray's recipe.  I'll make a new issue to track this: #157

aospy/test/test_automate.py
spencerahill: Yes this is a good point, and the same holds for `output_time_regional_reductions`.  The backend should handle this, rather than forcing the user to.

aospy/test/test_automate.py
spencerahill: Yes

aospy/automate.py
spencerahill: Yes

aospy/automate.py
spencerahill: Yes this looks like a good solution

(I actually inadvertently left this particular print statement from debugging.  But this will be a good solution for printing all the specs at once at the suite level)

spencerahill:
>Thanks for all of your work on this by the way!

Thank YOU.  Sorry this PR ended up so sprawling and for leaning on you so much.  Feel like I should have caught a lot of the issues you raised myself.

spencerkclark:
@spencerahill not a problem; it was good practice reviewing :)

I'm heading out for lunch now, but should be back online in an hour or so.  Feel free to make the changes I suggested in the meantime (just be sure to push updates frequently, so I know what you've done), otherwise I'll try to get to them when I get back

spencerahill:
>Feel free to make the changes I suggested in the meantime (just be sure to push updates frequently, so I know what you've done), otherwise I'll try to get to them when I get back

Yes, will do and will ping you

aospy/test/test_automate.py
spencerahill: Done in 93c7045

spencerahill:
@spencerkclark I'm still working through these, I'll ping you when I'm done, don't worry about it in the meantime (was going to push that same fix shortly)

spencerahill:
@spencerkclark ok I think I've addressed everything at this point, less some further tests of automate to bring the coverage up.

However see my note in the dd25efb commit re: the weird failing test.  On Travis it's only failing for 2.7, but it's not simply a Python version issue; it happens for me on 3.5 sometimes, but not always.

spencerkclark:
I want to say in this test we are implicitly assuming that the order is consistent across the sets that go into producing the dictionaries (which is not guaranteed).  Using sets is nice to automatically trim duplicates, but I wonder if we should just switch to lists for test simplicity, which always have the same order?

aospy/utils/io.py
```diff
@@ -38,17 +35,6 @@ def get_parent_attr(obj, attr, strict=False):
...
-    assert isinstance(objs, (tuple, list, dict, set))
-    if isinstance(objs, (tuple, list, set)):
-        try:
-            return {obj.name: obj for obj in objs}
-        except AttributeError as e:
-            raise AttributeError(e)
-    return objs
-
-
 def to_dup_list(x, n, single_to_list=True):
```
spencerkclark: I just realized we no longer use this (untested) function, so I think we can remove it

spencerahill: Good call, will delete

spencerahill:
>I wonder if we should just switch to lists for test simplicity, which always have the same order?

Yes this is probably the right move.  I was motivated by a vague desire to think more carefully about what data structures we are using in each place, but I guess in this case a list's order preservation is more useful than set's prevention of duplicates (there [is no built-in ordered set](http://stackoverflow.com/questions/1653970/does-python-have-an-ordered-set))

Will do this shortly

aospy/utils/io.py
spencerahill: FYI I recently learned of the [vulture](https://github.com/myint/vulture) package that finds unused code in a package, could be useful to use this periodically on aospy

aospy/test/test_automate.py
```diff
@@ -183,48 +182,31 @@ def test_permute_core_specs(self, calc_suite):
...
-            # set() only accepts hashables
-            except TypeError:
-                return False
-            else:
-                if not set_equal:
-                    return False
-    return True
+        assert len(actual) == len(expected)
+        for act in actual:
+            assert act in expected
```
spencerkclark: But of course!  Much cleaner solution here.  Sorry for leading you down the wrong path.

spencerahill:
Wow this ended up being a headache!

@spencerkclark automate is only at 78% coverage, but if it's alright with you I'd like to let it slide.  Partly because I see a clear path forward for an improved long-term solution (#158 , with more details to come).  Partly because I'm burnt out on this module!  Self-inflicted 😄

spencerahill:
@spencerkclark good catch on the docstring.  Just updated the what's new.

Since all CI passed on the previous non-doc commits, we're green in my view.  Any final concerns?

spencerkclark:
@spencerahill thanks again!  Bear with me, let me see if I can write a simple test for ```submit_mult_calcs``` and ```exec_calcs``` (it would just be to submit a single Calc, under both parallel and serial methods).

spencerkclark:
@spencerahill let me know if you're good with the test I added; if that checks out, and the AppVeyor build succeeds, I'm good to merge!

spencerahill:
@spencerkclark done!  Once again, huge thanks for catching my myriad mistakes on this and all your other effort on it.

Ironic that we went for a "quick fix", since it ended up being such an ordeal!  Just how it goes sometimes I guess 😄

PR #164, opened 2017-03-29
Improve docs tutorial

### LINKED ISSUE #151: Add example_obj_lib.py file using objects from tutorial.ipynb and use in main script

spencerahill:
Closes #151 

@spencerkclark with apologies for poor git practices, I ran out of steam before I could finish the last chunk of the Examples page (I noted the point where I stopped).

Will return to this tomorrow.  I feel like it's mostly there, but of course any suggestions welcome.


aospy/automate.py
```diff
@@ -290,34 +290,126 @@ def _print_suite_summary(calc_suite_specs):
...
+
+              - 'reg.av', 'reg.std', 'reg.ts' : analogous to 'av', 'std', 'ts'
+
+        output_vertical_reductions : {None, 'vert_av', 'vert_int'}, optional
+            How to reduce the data vertically:
+
+            - None : no vertical reduction
+            - 'vert_av' : mass-weighted vertical average
+            - 'vert_int' : mass-weighted vertical integral
+        input_time_intervals : {'annual', 'monthly', 'daily', '6hr', '3hr'}
```
spencerkclark: Maybe put an "e.g." here to note that these are not the only options (and that the values one needs to specify here are controlled by how one sets up their DataLoaders).

aospy/automate.py
```diff
@@ -290,34 +290,126 @@ def _print_suite_summary(calc_suite_specs):
...
-        - write_to_tar : If True, write results of calculations to .tar files,
-              one for each object.  These tar files have an identical directory
-              structures the standard output relative to their root directory,
-              which is specified via the `tar_direc_out` argument of each Proj
-              object's instantiation.
+        - prompt_verify : (default False) If True, print summary of
+              calculations to be performed and prompt user to confirm before
+              submitting for execution.
+        - parallelize : (default False) If True, submit calculations in
+              parallel.
```
spencerkclark: Should we note here that the way this is set up currently, this requires installing an optional dependency (```pip install multiprocess```)?

aospy/automate.py
```diff
@@ -290,34 +290,126 @@ def _print_suite_summary(calc_suite_specs):
...
+        - write_to_tar : (default True) If True, write results of calculations
+              to .tar files, one for each object.  These tar files have an
```
spencerkclark: "one for each object" feels somewhat ambiguous here; I think you mean "one for each :py:class`aospy.Run`"?

aospy/automate.py
```diff
@@ -290,34 +290,126 @@ def _print_suite_summary(calc_suite_specs):
...
+              identical directory structures the standard output relative to
```
spencerkclark: This is possibly a bug, but I'll just note that this is not entirely true at the moment.  For example the example main script, `aospy/examples/aospy_main.py`, produces the following directory structures:

Normal output:
`example-output/example_proj/example_model/example_run`

Tar output (note the `data` directory under `example_proj`):
`example-tar-output/example_proj/data/example_model/example_run`

docs/using-aospy.rst
```diff
@@ -239,7 +233,8 @@ and in the directory structure within which they are saved.
 - File name :
   ``varname.intvl_out.dtype_out_time.'from_'intvl_in'_'dtype_in_time.model.run.date_range.nc``
 
-See the API reference documentation of ``CalcInterface`` for explanation of each of these components of the path and file name.
+See the API reference documentation of ``CalcInterface`` for
```
spencerkclark: Should you use :py:class:`aospy.CalcInterface` here?

docs/using-aospy.rst
```diff
@@ -172,11 +167,10 @@ element.  E.g. ``models = ['name-of-my-model']``.
...
 .. note::
 
-   Although the main script is the recommended way to perform
-   calculations, it's possible to submit calculations by other means.
-   For example, one could explicitly create ``Calc`` objects and call
-   their ``compute`` method, as is done in the example Jupyter
-   notebook.
+   The main script is the recommended way to perform calculations.
+   Nevertheless, it's possible to submit calculations by other means,
+   such as by explicitly creating ``Calc`` objects and calling their
```
spencerkclark: Should you use :py:class:`Calc` here?

docs/using-aospy.rst
```diff
@@ -249,9 +244,10 @@ a ``CalcInterface`` object.  This object, in turn, is used to
...
 
-Unlike ``Proj``, ``Model``, ``Run``, ``Var``, and ``Region``, these
+Unlike :py:class:`aospy.Proj`, :py:class:`aospy.Model`,
+:py:class:`aospy.Run`, :py:class:`aospy.Var`, and ``Region``, these
 objects are not intended to be saved in ``.py`` files for continual
 re-use.  Instead, they are generated as needed, perform their desired
 tasks, and then go away.
 
-See the API reference documentation for further details.
+See the :ref:`API refeerence <api-ref>` documentation for further details.
```
spencerkclark: Typo: "refeerence"

docs/examples.rst
```diff
@@ -1,176 +1,264 @@
...
-from an idealized aquaplanet climate model.  A simple computation one
-could seek to do from this model output would be to compute some
-statistics of the total precipitation rate (large-scale plus
-convective).
+.. note::
+
+   The footnotes in this section provide scientific background to help
+   you understand the motivation and physical meaning of these example
+   calculations.  They can be skipped if you are familiar already or
+   don't care.
```
spencerkclark: This wording is a little bit harsh; maybe change "don't care" to "are not interested in those details."

docs/examples.rst
```diff
@@ -1,176 +1,264 @@
...
+
+These files contain timeseries of monthly averages of two variables
+generated by an idealized aquaplanet climate model: [#idealized]_
+[#aquaplanet]_
+
+1. Precipitation generated through gridbox-scale condensation
+2. Precipitation generated through the model's convective
+   parameterization [#var-defs]_
+
+Using this data that was directly outputted by our model, let's
```
spencerkclark: nit: I would refer to it as "the model" rather than "our model" here (I don't want to give the impression that we played a role in developing it), but it's totally up to you.

docs/examples.rst
```diff
@@ -1,176 +1,264 @@
...
+dictionary:
 
 .. ipython:: python
 
-    from datetime import datetime
-    
-    from aospy import Run
     from aospy.data_loader import DictDataLoader
-    file_map = {'monthly': '../aospy/test/data/netcdf/000[4-6]0101.precip_monthly.nc'}
+    file_map = {'monthly': rootdir + '/aospy/test/data/netcdf/000[4-6]0101.precip_monthly.nc'}
```
spencerkclark: Should we use `os.path.join``` here?

docs/examples.rst
```diff
@@ -1,176 +1,264 @@
...
 
 .. ipython:: python
 
     from aospy import Model
     example_model = Model(
         name='example_model',
         grid_file_paths=(
-            '../aospy/test/data/netcdf/00040101.precip_monthly.nc',
-            '../aospy/test/data/netcdf/im.landmask.nc'
+            rootdir + '/aospy/test/data/netcdf/00040101.precip_monthly.nc',
```
spencerkclark: Similar to above -- in these two lines should we use `os.path.join`?

docs/examples.rst
```diff
@@ -1,176 +1,264 @@
...
+            rootdir + '/aospy/test/data/netcdf/im.landmask.nc'
```
spencerkclark: It might be confusing to users why we are specifying a land mask for an aquaplanet simulation; should we make a note about this requirement?  In this case it **should**<sup>1</sup> be all zeros at all latitudes and longitudes.

<sup>1</sup>Good thing I checked...it turns out that it isn't.  This is a land mask for a rectangular continent.  I will submit a PR to correct this.

docs/examples.rst
```diff
@@ -183,42 +271,203 @@ infrastructure defined above, and also define an
...
+
+Each :py:class:`aospy.Calc` object includes the paths to the output
+and the results of each output type.
+
+.. ipython:: python
+
+    calcs[0].path_out
+    calcs[0].data_out
+
+**(S. Hill: Still need to finish up everything below this point)**
```
spencerkclark: Sorry, I recognize you're not done here, but I'm adding some comments anyway :)

docs/examples.rst
```diff
@@ -183,42 +271,203 @@ infrastructure defined above, and also define an
...
+
+.. note::
+
+   You may have noticed that ``subset_`` and ``raw_`` coordinates have
+   years 1678 and later, when our data was from model years 4
+   through 6.  This is because technical details upstream (in numpy)
```
spencerkclark: This issue is complicated and there are multiple levels of upstream dependencies; instead of calling out numpy in particular, I might just link to the discussion in this issue: https://github.com/pandas-dev/pandas/issues/7307

docs/examples.rst
```diff
@@ -183,42 +271,203 @@ infrastructure defined above, and also define an
...
+   limit the range of supported years to roughly 1677 to 2234.
```
spencerkclark: Let's link directly to this part of the pandas docs here: http://pandas.pydata.org/pandas-docs/stable/timeseries.html#timestamp-limitations

Also the upper bound year should be 2262 (instead of 2234).

docs/examples.rst
```diff
@@ -183,42 +271,203 @@ infrastructure defined above, and also define an
...
+precipitation is dominated by the convective rainfall, and the
+tropical mean even more so.
+
+Beyond this simple example
+--------------------------
+
+We can re-use our object library at will to perform new calculations
+or re-compute old ones.  We can also add new objects.  For example,
+suppose we performed a new simulation in which we increased the
+surface albedo or introduced a rectangular continent.  All we would
```
spencerkclark: In this case of a rectangular continent, I might say "not so fast!" :)

Currently the safest thing to do in that situation would be to create a new Model with a different land mask (to make sure regional calculations that depended on land would be accurate).  Changing surface albedo is fine for the same Model, however.  

If you want some other parameters to tweak as an example (while keeping the same Model object) you could say increase the solar constant or turn off the convection scheme.

spencerkclark (commented):
@spencerahill this looks very good!  Thanks for all your effort here cleaning up the internal links in the docs, fleshing out the ```submit_mult_calcs``` docstring, and greatly improving the examples page by adding more description of the aospy workflow, and porting it to mirror (and ultimately use) the ```example_obj_lib```.  Adding the scientific details as footnotes is also a really nice touch.

I've gone through your changes so far and offered some comments/suggestions; hopefully you find them helpful.

docs/examples.rst
spencerkclark: See #165 for the fix

spencerahill:
@spencerkclark thanks a lot!  Won't bother responding in-line, because it all sounds good.  Will merge #165 also.

aospy/automate.py
spencerahill: Good catch!  I agree, this is a bug in my view.  I will fix it within this PR if you agree.

spencerkclark: Go for it!

spencerahill:
@spencerkclark ok I've addressed all your comments and finished the Examples section.

I will go through all of the docs pages later today one more time...I suspect there are some out of date parts in Using aospy especially, given all the recent modifications.

After that I'd say ready to merge.

spencerkclark:
This looks great to me; go ahead and merge!  Thanks again.  

> I will go through all of the docs pages later today one more time...I suspect there are some out of date parts in Using aospy especially, given all the recent modifications.

Sounds good; I'll take care of updating the example Jupyter notebook.  I think I should be able to get our CI configured to test that as well.

spencerahill:
Ok, finished going through all the docs again.  Merging pending CI pass

docs/using-aospy.rst
```diff
@@ -233,21 +209,49 @@ and in the directory structure within which they are saved.
...
-Unlike :py:class:`aospy.Proj`, :py:class:`aospy.Model`,
-:py:class:`aospy.Run`, :py:class:`aospy.Var`, and
-:py:class:`aospy.Region`, these objects are not intended to be saved
-in ``.py`` files for continual re-use.  Instead, they are generated as
-needed, perform their desired tasks, and then go away.
+==============
+
+:py:func:`aospy.submit_mult_calcs` creates a :py:class:`aospy.CalcSuite`
+object that permutes over the provided lists of calculation
+specifications, encoding each permutations into
```
spencerkclark: Typo: "each permutations"

spencerkclark (commented):
Looks good!  Just caught a minor typo.

docs/using-aospy.rst
spencerahill: good catch, thanks

PR #166, opened 2017-03-29
WIP Update and test notebook

spencerkclark:
The current push updates the notebook just enough so that it runs successfully.  I think ideally I should update it to mirror the new docs.  I need to head out now, but I should have some time this evening to get to that.


spencerahill:
Cool!  Interesting that only the AppVeyor build failed...will let you take the initial stab at that, but let me know if there's something I can do

spencerkclark:
Yes, I'll have a look into the AppVeyor issue.

As far as content goes, xarray's example notebooks exactly mirror their doc pages (I think they use a tool to create doc pages from notebooks).  I don't suggest we do the same thing here (meaning use a tool, because that limits the formatting options in the docs, which I think are really nice right now), but as far as textual descriptions go, should I copy-paste from your descriptions or should I try to trim things down a bit?

spencerkclark:
> copy-paste from your descriptions 

I went ahead with this route to mimic xarray.  Please let me know if you'd like me to make any changes.

Still trying to sort out the AppVeyor issue -- will return to this tomorrow morning.

spencerahill:
>I went ahead with this route to mimic xarray. Please let me know if you'd like me to make any changes.

Yes, that seems best for now.

spencerkclark:
b13ec73 confirms that the issue on Windows stems from using a Unix-style symbolic link to point to the tutorial notebook.  [It doesn't look like there's a good way to reconcile this to be compatible with both systems.](http://unix.stackexchange.com/questions/63172/does-windows-recognize-linuxs-symbolic-links)

To be fair, what we are doing is not entirely the intended use-case of pytest-ipynb.  It is actually meant for real unit-tests defined in notebook form, not simply executing all the cells of a notebook to see if they run.  This is why it requires for collection purposes that all notebooks it will run as tests have names of the form `test*.ipynb`, hence the hack to create a symbolic link to the tutorial notebook with the name `test_tutorial.ipynb` rather than rename the notebook itself (this hack was borrowed from [OceanPARCELS/parcles](https://github.com/OceanPARCELS/parcels/tree/master/examples)).

**Options for using pytest-ipynb:**

While pytest-ipynb is a little nicer for its ability to run cells even if preceding cells raise errors, and for its better integration with pytest (treating each cell as a separate unit test), having to resort to these hacks is a little messy.  That said if we want to use pytest-ipynb, there would be two options:
1.  Make a copy of `tutorial.ipynb` called `test_tutorial.ipynb`; this has the downside that we always need to remember to make a copy whenever we update the notebook.
2.  Use a symbolic link, but don't test the notebook on Windows (i.e. create separate CI environments that don't include pytest-ipynb on Windows; therefore the tests from the symlink wouldn't be collected and run by pytest there); this has the downside that we wouldn't catch Windows-specific errors (this is particularly important for catching issues with filepaths, though using the `os` module for all file operations tends to prevent that).

**Alternative to pytest-ipynb:**

An alternative is to not use pytest-ipynb (my latest push).  For our purposes, since we just want a quick check to make sure *all* the cells of the notebook execute without error, we can actually make do without pytest-ipynb and create a file (called `test_tutorial.py`) in the `test` directory that looks like:

```
import pytest

from IPython.nbformat.current import read
from runipy.notebook_runner import NotebookRunner


def test_tutorial_notebook():
    with open('/path/to/tutorial.ipynb') as nb:
        notebook = read(nb, 'json')
    r = NotebookRunner(notebook)
    r.run_notebook()
```

This just defines a test function that gets collected by conventional pytest, that reads in and runs all the cells of the tutorial notebook.  If it encounters an exception while running the notebook, the test will fail and the traceback will be messy, but reasonably informative (see below).  This has two advantages:
1.   It runs on all operating systems
2.  It does not require using symbolic links or making copies of notebooks

@spencerahill what are your thoughts here?  It also occurs to me that we should include some logic to skip tests that require other libraries for testing, but are not strictly required by aospy (so users can run the test suite on their machines without getting import errors).  That's probably for a separate PR, however (see [xarray's `tests/__init__.py`](https://github.com/pydata/xarray/blob/master/xarray/tests/__init__.py) file for some examples of this kind of machinery).

```
(aospy-dev) bash-4.1$ pytest test_tutorial.py
======================================================== test session starts =========================================================
platform linux -- Python 3.6.0, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /home/skc/aospy, inifile: setup.cfg
plugins: ipynb-1.1.0, cov-2.4.0
collected 1 items

test_tutorial.py F

============================================================== FAILURES ==============================================================
_______________________________________________________ test_tutorial_notebook _______________________________________________________

    def test_tutorial_notebook():
        with open('../examples/tutorial.ipynb') as nb:
            notebook = read(nb, 'json')
        r = NotebookRunner(notebook)
>       r.run_notebook()

test_tutorial.py:12:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../miniconda/envs/aospy-dev/lib/python3.6/site-packages/runipy/notebook_runner.py:232: in run_notebook
    self.run_cell(cell)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <runipy.notebook_runner.NotebookRunner object at 0x7f6b5ef28550>
cell = {'cell_type': 'code', 'metadata': {'deletable': True, 'editable': True}, 'outputs': [{'output_type': 'pyerr', 'ename':...r, '000[4-6]0101.precip_monthly.nc3'),\n                  decode_times=False)", 'prompt_number': 2, 'collapsed': False}

    def run_cell(self, cell):
        """Run a notebook cell and update the output of that cell in-place."""
        logging.info('Running cell:\n%s\n', cell.input)
        self.kc.execute(cell.input)
        reply = self.kc.get_shell_msg()
        status = reply['content']['status']
        traceback_text = ''
        if status == 'error':
            traceback_text = 'Cell raised uncaught exception: \n' + \
                '\n'.join(reply['content']['traceback'])
            logging.info(traceback_text)
        else:
            logging.info('Cell returned')

        outs = list()
        while True:
            try:
                msg = self.kc.get_iopub_msg(timeout=1)
                if msg['msg_type'] == 'status':
                    if msg['content']['execution_state'] == 'idle':
                        break
            except Empty:
                # execution state should return to idle
                # before the queue becomes empty,
                # if it doesn't, something bad has happened
                raise

            content = msg['content']
            msg_type = msg['msg_type']

            # IPython 3.0.0-dev writes pyerr/pyout in the notebook format
            # but uses error/execute_result in the message spec. This does the
            # translation needed for tests to pass with IPython 3.0.0-dev
            notebook3_format_conversions = {
                'error': 'pyerr',
                'execute_result': 'pyout'
            }
            msg_type = notebook3_format_conversions.get(msg_type, msg_type)

            out = NotebookNode(output_type=msg_type)

            if 'execution_count' in content:
                cell['prompt_number'] = content['execution_count']
                out.prompt_number = content['execution_count']

            if msg_type in ('status', 'pyin', 'execute_input'):
                continue
            elif msg_type == 'stream':
                out.stream = content['name']
                # in msgspec 5, this is name, text
                # in msgspec 4, this is name, data
                if 'text' in content:
                    out.text = content['text']
                else:
                    out.text = content['data']
            elif msg_type in ('display_data', 'pyout'):
                for mime, data in content['data'].items():
                    try:
                        attr = self.MIME_MAP[mime]
                    except KeyError:
                        raise NotImplementedError(
                            'unhandled mime type: %s' % mime
                        )

                    setattr(out, attr, data)
            elif msg_type == 'pyerr':
                out.ename = content['ename']
                out.evalue = content['evalue']
                out.traceback = content['traceback']
            elif msg_type == 'clear_output':
                outs = list()
                continue
            else:
                raise NotImplementedError(
                    'unhandled iopub message: %s' % msg_type
                )
            outs.append(out)
        cell['outputs'] = outs

        if status == 'error':
>           raise NotebookError(traceback_text)
E           runipy.notebook_runner.NotebookError: Cell raised uncaught exception:
E           ---------------------------------------------------------------------------
E           OSError                                   Traceback (most recent call last)
E           <ipython-input-2-46ac98b26f33> in <module>()
E                 1 import xarray as xr
E                 2 xr.open_mfdataset(os.path.join(rootdir, '000[4-6]0101.precip_monthly.nc3'),
E           ----> 3                   decode_times=False)
E
E           /home/skc/miniconda/envs/aospy-dev/lib/python3.6/site-packages/xarray/backends/api.py in open_mfdataset(paths, chunks, concat_dim, compat, preprocess, engine, lock, **kwargs)
E               498         paths = sorted(glob(paths))
E               499     if not paths:
E           --> 500         raise IOError('no files to open')
E               501
E               502     if lock is None:
E
E           OSError: no files to open

../../../miniconda/envs/aospy-dev/lib/python3.6/site-packages/runipy/notebook_runner.py:214: NotebookError
-------------------------------------------------------- Captured stderr call --------------------------------------------------------
INFO:root:Running cell:
import os  # Python built-in package for working with the operating system
import aospy
rootdir = os.path.join(aospy.__path__[0], 'test', 'data', 'netcdf')

INFO:root:Cell returned
INFO:root:Running cell:
import xarray as xr
xr.open_mfdataset(os.path.join(rootdir, '000[4-6]0101.precip_monthly.nc3'),
                  decode_times=False)

INFO:root:Cell raised uncaught exception:
OSError                                   Traceback (most recent call last)
<ipython-input-2-46ac98b26f33> in <module>()
      1 import xarray as xr
      2 xr.open_mfdataset(os.path.join(rootdir, '000[4-6]0101.precip_monthly.nc3'),
----> 3                   decode_times=False)

/home/skc/miniconda/envs/aospy-dev/lib/python3.6/site-packages/xarray/backends/api.py in open_mfdataset(paths, chunks, concat_dim, compat, preprocess, engine, lock, **kwargs)
    498         paths = sorted(glob(paths))
    499     if not paths:
--> 500         raise IOError('no files to open')
    501
    502     if lock is None:

OSError: no files to open
====================================================== 1 failed in 4.57 seconds ======================================================
```

spencerkclark:
@spencerahill I'm a little puzzled by what's happening with the python 2.7 environment on Travis; the build log seems to indicate that it's done and everything passed, but for some reason the job hangs.

spencerahill:
>what we are doing is not entirely the intended use-case of pytest-ipynb. It is actually meant for real unit-tests defined in notebook form

I had thought that when I looked at their documentation, but I thought maybe there was a secondary mode/functionality for what we are after that I hadn't noticed.

>For our purposes, since we just want a quick check to make sure all the cells of the notebook execute without error, we can actually make do without pytest-ipynb and create a file (called test_tutorial.py) in the test directory

Yes, I think this is the route to go.  It's way less hacky, and, as discussed above, we were really trying to shoehorn pytest-ipynb into something it wasn't meant for.  Still a useful diversion though!

>we should include some logic to skip tests that require other libraries for testing

Yes, I agree, and mimicking xarray (as is usually the case) seems like the way to go.  Is it any more complicatd than the following?  If not, then let's just do it now for this test at least:

```python
try:
    import IPython
    has_ipython = True
except ImportError:
    has_ipython = False

try:
    import runipy
    has_runipy = True
except ImportError:
    has_runipy = False

def requires_ipython(test):
    return test if has_ipython else pytest.mark.skip('requires ipython')(test)

def requires_runipy(test):
    return test if has_runipy else pytest.mark.skip('requires runipy')(test)

...

@requires_ipython
@requires_runipy
def test_tutorial_notebook():
    ...
```

>puzzled by what's happening with the python 2.7 environment on Travis

Same, let's see if it happens again after your next commit(s)

spencerahill:
@spencerkclark we should also add IPython to the CI environment files

spencerkclark:
> Is it any more complicatd than the following?

Indeed that was exactly what I had in mind!  Hopefully python 2.7 on Travis will get closed this time around.  Let me know if you have any more comments on this PR.

spencerahill:
Great!

No further comments, will merge if CI is green.

spencerkclark:
Travis was hanging again -- giving this a try...

spencerahill:
Nice, looks like that did the trick

spencerahill:
FYI I'll be away from the computer for next ~2 hours, feel free to merge if AppVeyor passes in the meantime.  Otherwise I'll do later today.

spencerahill:
@spencerkclark thanks again for this!

PR #175, opened 2017-04-24
Remove unused operator.py module

### LINKED ISSUE #174: operator.py module name conflicts with a standard library module's
This is a somewhat obscure issue, but it came up today as I was updating my `emacs` configuration on the GFDL systems -- I now use [flycheck](http://www.flycheck.org/en/latest/) for doing syntax checking/linting of python, which depends on the [standard library's `operator` module](https://docs.python.org/3/library/operator.html).  When I edit files in the aospy repository, python tries to import aospy's version of the `operator` module, which causes problems for flycheck.

There are two simple solutions here:
1. Rename `aospy/operator.py` something else
2. Since we don't currently use any of the code in `aospy/operator.py`, we could remove it for now

@spencerahill what are your thoughts?  Should we try to fix this?

spencerkclark:
Closes #174


spencerahill:
Awesome, thanks!  Let's also remove [its section](https://github.com/spencerahill/aospy/blob/develop/docs/api.rst#operator) in the API reference docs page.  And I'd do a grep through the whole package after that for 'operator' to make sure there are no other dangling references.

Once you've done those, go ahead and merge.  This doesn't need a what's new entry.

spencerahill:
Actually let's just add a what's new entry under Bug fixes, which can just be 
```
- Remove obsolete operator.py module (fixes :issue:`174` via :pull:`175`).  By `Spencer Clark <https://github.com/spencerkclark>`_.
```

spencerkclark:
> Let's also remove its section in the API reference docs page.

Ah good catch -- I should have grepped at the package level rather than the core level.  I think we should be all set now!  I'll go ahead and merge once the CI finishes.

PR #176, opened 2017-04-26
Skip tests that use optional dependencies if not installed

### LINKED ISSUE #167: Add decorators to potentially skip tests that require optional dependencies
Currently we have tests which require:
- multiprocess

And if we add tests of the notebook they will also require:
- matplotlib
- runipy or pytest-ipynb

None of these are required dependencies for aospy, so we should only run those tests if those libraries are present on someone's system.  xarray has some good examples for how to do this in: [tests/__init__.py](https://github.com/pydata/xarray/blob/master/xarray/tests/__init__.py).

spencerahill:
Closes #167 

Still need to configure CI to use the new minimal dependencies environment.

On this pass I just used `pytest.importorskip`.  ~~I couldn't see any benefit of xarray's more involved route that I originally mimicked.  Any thoughts on this?~~ (see below)


spencerahill:
Actually for pytest-catchlog, `pytest.importorskip` isn't a viable solution.  Doing it outside the test causes all of the modules tests to be skipped.  Doing it inside the test doesn't work because the `caplog` fixture causes an error before `pytest_catchlog` can be imported.

So the solution is to mimic xarray's try/except-based decorator approach.

spencerahill:
The one failing Travis build is from an error on their side in the 3.4 env; so this should be ok pending Appveyor tests finishing.

@spencerkclark could you review this when you get a chance?  Thanks!

spencerkclark:
This looks good to me!  I went ahead and restarted the Travis build for 3.5, and it looks like things cleared up.  Thanks for taking care of this.

spencerkclark:
Actually, should this get a what's new entry?  I suppose one could consider this a "bug fix," considering that without this setup, the test suite wouldn't pass if one only had the minimum requirements installed.

spencerahill:
Thanks @spencerkclark!

PR #178, opened 2017-05-04
Remove multi-run logic from calc

### LINKED ISSUE #117: Remove legacy code in calc meant to enable calculations involving multiple runs and models
As part of #90 I will be making a first step towards a refactor of calc.py.  Before doing that, however, I think it might be a good idea to clean up some of the code that was at one point meant as a crude way to enable calculations involving multiple runs and models.  To my knowledge this functionality has been broken for some time, is not tested or documented, and what's left of the logic seems to muddy things a fair amount in calc.py.

@spencerahill what are your thoughts on this?  I know eventually we'll probably want to revisit this type of functionality, but for now do you think we should try for a start fresh?

spencerahill:
Closes #117 

Of course we do eventually want to re-implement this, but good in the meantime to get rid of this really ugly and confusing code.


aospy/calc.py
```diff
@@ -371,24 +340,22 @@ def _get_pressure_vals(self, var, start_date, end_date, n=0):
...
-            return self._get_pressure_from_eta_coords(ps, name=var.name, n=n)
+            return self._get_pressure_from_eta_coords(ps, name=var.name)
         raise ValueError("`dtype_in_vert` must be either 'pressure' or "
                          "'sigma' for pressure data")
 
-    def _get_input_data(self, var, start_date, end_date, n):
+    def _get_input_data(self, var, start_date, end_date):
         """Get the data for a single variable over the desired date range."""
         logging.info(self._print_verbose("Getting input data:", var))
         # If only 1 run, use it to load all data.  Otherwise assume that num
```
spencerkclark: Is this comment still relevant?

spencerkclark (commented):
Thanks for taking care of this cleanup!

I think the test failures may be related to https://github.com/dask/distributed/issues/1056

aospy/calc.py
spencerahill: Good call, will go through all the comments looking for anything else along these lines.

spencerahill:
Fixed that comment and snuck in some other minor cleanup in calc.py

>I think the test failures may be related to dask/distributed#1056

@spencerkclark I think you're right.  Funny that it's not happening for 3.4.  I'm comfortable merging as is despite those failures; let me know if you agree.  Tests are all passing on my local machine.

docs/whats-new.rst
```diff
@@ -30,6 +30,10 @@ Dependencies
 Bug Fixes
 ~~~~~~~~~
 
+- Remove faulty logic for calculations with data coming from multiple
+  runs.  Eventually this feature will be properly implemented (fixes
+  :issue:`###` via :pull:`17#`).  By `Spencer Hill
```
spencerkclark: Fill in issue numbers here (#117 and #178)

spencerkclark (commented):
Looks good to me pending the minor fix to the what's new page.  I agree that the test failures are unrelated;  I say go ahead and merge after filling in the issue numbers.

docs/whats-new.rst
spencerahill: Oops! another good catch

spencerahill:
Great, thanks!

PR #180, opened 2017-05-04
Add ability to specify custom preprocessing functions

### LINKED ISSUE #177: Add option for custom user preprocessing step to data_loader.load_variable?
We toyed with adding this in #90, but eventually settled on just adding a custom time-offset capability for simplicity.  

I know someone trying out aospy that would like to use it for looking at output from the WRF model, [which unfortunately does not always comply with CF conventions](http://foehn.colorado.edu/wrfout_to_cf/); this is particularly problematic for the time variable:

```
<xarray.DataArray 'XTIME' (Time: 360)>
array([ 259440.,  259680.,  259920., ...,  345120.,  345360.,  345600.], dtype=float32)
Dimensions without coordinates: Time
Attributes:
    FieldType:    104
    MemoryOrder:  0
    description:  minutes since simulation start
    units:
    stagger:
```
It would be nice, rather than have to modify every output file to have a CF-compliant time units variable, if a user could provide a function to apply to the dataset before aospy touches it.  I think it would make most sense for this to be a specification one could make as an optional argument in a DataLoader constructor.  

For instance:
```python
def fix_wrf_time_units(ds):
    if 'XTIME' in ds:
        ds['XTIME'].attrs['units'] = 'minutes since 0001-06-30 04:00:00'
    return ds

rootdir = '/archive/xrc/wrf_output/WTG_colcool/28C_1pt5_noT'

_file_map = {'4hr': os.path.join(rootdir, 'wrfout_d01_0001-06-30_04:00:00')}
example_run = Run(
    name='example_run',
    description=(
        'WRF WTG simulation'
    ),
    default_start_date=datetime.datetime(1, 6, 30) + datetime.timedelta(days=50),
    default_end_date=datetime.datetime(1, 6, 30) + datetime.timedelta(days=60),
    data_loader=DictDataLoader(_file_map, preprocess={'4hr': fix_wrf_time_units})
)
```
The `preprocess` argument could take a dictionary as input, with an `intvl_in` mapping to a user-defined function, which would be called before [`grid_attrs_to_aospy_names`](https://github.com/spencerahill/aospy/blob/develop/aospy/data_loader.py#L13) whenever files were loaded from a particular file set.  This could be used to clean up datasets that are close, but not quite, compatible with aospy's assumptions.

@spencerahill what are your thoughts here?  Does this sound like a decent option to address this issue?

spencerkclark:
Closes #177

@spencerahill this needs better documentation and eventually a what's new, but let me know your thoughts on the initial approach.


spencerahill:
Thanks for this.  My one initial question is: can we make use of the `apply_preload_user_commands` logic we already have for this?  It seems like this is in line with what we had envisioned it for.

I'll look through more closely again once we sort that out.

spencerkclark:
> My one initial question is: can we make use of the apply_preload_user_commands logic we already have for this?

The issue is the commands called by `apply_preload_user_commands` operate on file paths (not xarray objects); my intent was for users to be able to treat the preprocessing functions enabled by this PR the same way they would treat functions one could pass as a `preprocess` argument to [`xr.open_mfdataset`](http://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html?).

aospy/data_loader.py
```diff
@@ -10,6 +10,26 @@
 from .utils import times, io
 
 
+def preprocess_and_rename_grid_attrs(func):
```
spencerahill: This won't be public API, right?  So add leading underscore.

spencerahill: Also maybe add 1 sentence to doc explaining why this is necessary, namely to generate single function that gets passed to 'preprocess' within open_mfdataset.  EDIT: and one sentence describing why it might be used, e.g. the WRF example.

spencerahill: Convention is to start w/ single line summary.  You don't need to specify the 'grid_atts_to_aospy_names` name explicitly, since that takes a lot of space

aospy/data_loader.py
```diff
@@ -10,6 +10,26 @@
...
+    Parameters
+    ----------
+    func : function
+       An arbitrary function to call before calling
+       ``grid_attrs_to_aospy_names`` in ``_load_data_from_disk``.  Must take
+       only an xr.Dataset as an argument.
+    
+    Returns
+    -------
+    function
```
spencerahill: Describe.  Can be essentially same language as in extra sentence I ask for above.

aospy/data_loader.py
```diff
@@ -248,9 +288,10 @@ class DictDataLoader(DataLoader):
     ...             '3hr': '000[4-6]0101.atmos_8xday.nc'}
     >>> data_loader = DictDataLoader(file_map)
     """
-    def __init__(self, file_map=None):
+    def __init__(self, file_map=None, preprocess_funcs=None):
```
spencerahill: So we're only implementing it for DictDataLoader?

spencerahill:
@spencerkclark ok that's all from me for now

spencerkclark:
@spencerahill I'm still not completely happy with this, but I need to run -- will add some comments later regarding my concerns.

spencerkclark:
So in fb09cc3 I've basically just copied the behavior of the `intvl_in`-dependent preprocess function to the other DataLoaders (`NestedDictDataLoader` and `GFDLDataLoader`).  In thinking about it more, it feels a little clunky to ask users to specify multiple functions based on the value of `intvl_in`.  In the example of the WRF use-case it might simply mean returning `days since 0001-06-30` versus `minutes since 0001-06-30`; it doesn't really make sense to define separate hard-coded functions to do this, rather it makes more sense to have one function that takes `intvl_in` as an argument and decides what to do to the `Dataset` based on that.  It's also not very flexible (what if users their preprocessing step to depend on different attributes?).

Perhaps an option could instead be to allow users to define a *single* preprocess function, which would take all the "DataLoader attributes" passed to `data_loader.load_variable` in `calc.py` (a list can be found in the [`_generate_file_set` function header](https://github.com/spencerahill/aospy/blob/develop/aospy/data_loader.py#L255)).   This would extend easily to all DataLoaders and would allow for as much flexibility as possible; the one drawback I see is the amount of boilerplate passed on to the user associated with requiring them to enumerate all the input arguments (even if they are not used by their particular preprocess function), though maybe there is a way to avoid this?

Does that make sense?  What do you think about that alternative?

spencerahill:
>it feels a little clunky to ask users to specify multiple functions based on the value of intvl_in. In the example of the WRF use-case it might simply mean returning days since 0001-06-30 versus minutes since 0001-06-30; it doesn't really make sense to define separate hard-coded functions to do this, rather it makes more sense to have one function that takes intvl_in as an argument and decides what to do to the Dataset based on that. It's also not very flexible (what if users their preprocessing step to depend on different attributes?).

I agree with all this.

>one drawback I see is the amount of boilerplate passed on to the user associated with requiring them to enumerate all the input arguments (even if they are not used by their particular preprocess function), though maybe there is a way to avoid this?

Can we use `**kwargs` here?  E.g. if only `intvl_in` was to be used: `def my_preprocess_func(intvl_in=None, **kwargs)`?  In that case, `**kwargs` would catch all of the unused keyword arguments.

In fact could we even automatically add this behind the scenes, so they're not even responsible for the `**kwargs` signature?  But I think this requires them passing their function arguments anyways.  E.g.

```python
def _add_kwargs_to_user_func(func, *funcargs):  # funcargs 
    def func_with_kwargs(func, *funcargs, **kwargs):
        return func(*funcargs)
    return func_with_kwargs
```

Not sure if that's clear; let me know if not.

spencerkclark:
> Can we use **kwargs here?

Indeed, I think something along the lines of your example could work well.  I'll give it a try.

spencerkclark:
@spencerahill there are some complications associated with trying to be too clever with adding `**kwargs` ourselves to a user's function (namely since we always pass a complete set of keyword arguments down from `load_variable` python will complain if one defines a default value for an attribute in their own function, because it will have been specified twice in the function call).  

What I've settled on here is having users specify functions of the form `def preprocess(ds, **kwargs)`.  This is still perhaps not ideal, but I'm having trouble thinking of a better solution.  I'm aware of [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial), but I don't think we can make use of it here since we don't know a priori what arguments the user would want to override in their preprocessing functions.

aospy/data_loader.py
```diff
@@ -247,10 +285,27 @@ class DictDataLoader(DataLoader):
     >>> file_map = {'monthly': '000[4-6]0101.atmos_month.nc',
     ...             '3hr': '000[4-6]0101.atmos_8xday.nc'}
     >>> data_loader = DictDataLoader(file_map)
+
+    If one wanted correct a CF-incompliant units attribute on each Dataset read
```
spencerahill: typo

spencerahill:
>What I've settled on here is having users specify functions of the form def preprocess(ds, **kwargs). This is still perhaps not ideal, but I'm having trouble thinking of a better solution.

I think this is a perfectly legitimate solution.  Anything beyond is probably getting carried away.

This looks good to me besides the typo.  Thanks for adding the example to the docstring, that's useful.  Need a what's new.

spencerkclark:
Thanks!  Have another look at additional the doc edits I made, but I think it might be ready to go now.

spencerahill:
Great, thank @spencerkclark!

spencerahill:
That's weird, I just got AppVeyor notification of test fail for 3.5: https://ci.appveyor.com/project/spencerahill/aospy/build/1.0.257/job/k5wr2pfdtc06slsb

@spencerkclark any ideas what's going on there?  Weird b/c AppVeyor shows green on this PR page.

spencerkclark:
I've found (it seems more so lately than before for some reason) that writing to a tar file can be unreliable in parallel (sometimes it works sometimes it doesn't).  I think it might have to do with the fact that two processes are trying to write to it at the same time (see also this failure: https://travis-ci.org/spencerahill/aospy/jobs/230198469).

I think whenever you merge a PR the CI suite runs again to update the badges on the repo home page, so it's possible that the test passed when I pushed to the PR, but failed after you merged it.  So long story short, I think it's unrelated to this PR, but something we probably should think about addressing in some way.

spencerahill:
Thanks, that all makes sense.  I agree, let's definitely keep our eye on it.  If it comes up again at all, we should at the very least add a warning in the docs about it, and simultaneously start thinking about a fix.

Issue #182, opened 2017-05-08
Support user-specified internal names of grid attributes


spencerahill:
Users can currently specify the names of the data variables in their input data via the `name` and `alt_names` attributes of `Var`.  But the same can't be done for grid attributes such as `lat`, `land_mask`, etc., because these are taken directly from the `GRID_ATTRS` dict in internal_names.py.

This effectively means that if their data's grid attributes don't match any of the names in `GRID_ATTRS`, they will be unable to use aospy unless they modify the aospy source code directly.  This just came up for me in trying to use some NCAR CAM output, which has 'nbnd' as its 'bounds' name and 'LANDFRAC' as its 'land_mask' name.

One initial idea (though more though/discussion definitely warranted): create `Var` objects within aospy for each grid attribute, using the `alt_names` attribute to store what's currently in GRID_ATTRS.  And then the user can define these Vars separately if they want, and aospy will look for them first in their object library before looking at the internal "master" list.

Or, perhaps less boilerplate, do the same but just look for a 'GRID_ATTRS' attribute in their object library.  I think that might be better; it's definitely easier.

I'm a bit all over the place right now... @spencerkclark am I right about the overall premise of this issue, or have I forgotten a way to do this?

spencerkclark:
+1 for raising this issue; incidentally this would also help the WRF users (e.g. `'lat'` and `'lon'` are named `'XLAT'` and `'XLONG'` respectively).  

> Or, perhaps less boilerplate, do the same but just look for a 'GRID_ATTRS' attribute in their object library. I think that might be better; it's definitely easier.

I would lean towards this solution as well; it seems like it would be relatively clean to implement, and would be relatively simple from a user's perspective.

spencerahill:
>this would also help the WRF users (e.g. 'lat' and 'lon' are named 'XLAT' and 'XLONG' respectively).

Good to know -- let me know all other of WRF's grid attributes that are non-standard, and I'll add them in the same PR

>I would lean towards this solution as wel

Great, we'll go this route.

spencerkclark:
> let me know all other of WRF's grid attributes that are non-standard, and I'll add them in the same PR

Do we need to hard-code the settable attributes or could we just have a user specify their own dictionary with only the attributes they want to add alternative names to, and then just iterate over the keys and values of that dictionary and update the internal `GRID_ATTRS` dictionary where specified?

spencerahill:
>Do we need to hard-code the settable attributes

[...]

Issue #80, opened 2016-07-05
Create abstract VertCoord class that all vertical coordinates inherit from


spencerahill:
(copied from offline convo w/ me and @spencerkclark in May)

Maybe it makes sense to create an abstract class or interface for vertical coordinates...it would have attributes and/or methods for layer interface values and layer center values.  Then users could specify their own coordinates, and aospy can use them so long as they adhere to the interface.  I'm sure there are other coordinate systems in use out there, especially on the ocean (or non-atmosphere more broadly) side than the pressure and hybrid sigma-pressure ones we use.  And that would contribute to our broader effort to refactor the codebase to be more modular.

infinite-diff already tries to do something similar.  And [Iris has something of this sort implemented](http://scitools.org.uk/iris/docs/latest/iris/iris/aux_factory.html).

spencerahill:
The need for something like this, or more generally a re-vamping of how we handle vertical coordinates, is now more acute, due to new use cases from different models, each with their own idiosyncrasies.  See #184 for the WRF model and #191 for the CAM model.

Another use case is sigma coordinates, i.e. pressure divided by surface pressure (confusingly we use the 'sigma' label to refer to hybrid sigma-pressure coordinates).  @spencerkclark it just occurred to me that the idealized models use this, including the test data we package with aospy!  How are we handling these coordinates right now?

We also need to provide a public API for users to define and supply their own vertical coordinate systems, so that they don't find themselves unable to use aospy if their data uses a vertical coordinate system that aospy doesn't support.

What would this look like?  The only things that I think are ultimately needed for each coordinate type are methods for computing the pressure at level interfaces and level centers.  (Although that pre-supposes that the vertical coordinate is ultimately pressure-related, which isn't the case for e.g. ocean models or atmospheric models that use a z coordinate).

spencerkclark:
> How are we handling these coordinates right now?

The FMS idealized models still output `bk` and `pk` variables; it just happens that `pk` is zero for all levels (so in that sense the pressure can be treated as being in hybrid coordinates since sigma coordinates are just a limiting case). I agree this is confusing though.

spencerahill:
>The FMS idealized models still output bk and pk variables; it just happens that pk is zero for all levels (so in that sense the pressure can be treated as being in hybrid coordinates since sigma coordinates are just a limiting case).

[...]

PR #183, opened 2017-05-09
Add CAM and WRF internal names for grid attributes

spencerahill:
C.f. discussion in #182.  Doesn't close that issue though -- will do the `Model.grid_attr` and `data_loader` aspects in a separate PR.


spencerkclark:
@spencerahill thanks for putting this together so quickly!  If you don't mind, could we delay merging a bit?  I've inquired offline about vertical coordinates in WRF (still waiting for a response), which may lead to some additions to this PR.

spencerahill:
Sure no rush, just wanted to get it out while it was fresh on my mind.  Just ping whenever you're ready.

spencerkclark:
@spencerahill sorry for taking a bit to get back to you on this.  I say go ahead and merge these updates for now.  

I think we'll need to think about how to handle vertical coordinates in WRF separately -- one can compute pressure at the level midpoints as the sum of a "perturbation pressure" and a constant.  The "perturbation pressure" is a fully 4D variable (varying in time, lat, lon, and the vertical).  So it doesn't cleanly fit in with the current options supported ('sigma' and 'pressure').  This possibly puts a little more incentive on resolving #80, and having a bit more of a general framework for handling vertical coordinates (outside the scope of this PR).

spencerahill:
Thanks for diving into this.

>one can compute pressure at the level midpoints as the sum of a "perturbation pressure" and a constant. The "perturbation pressure" is a fully 4D variable (varying in time, lat, lon, and the vertical).

Interesting.  So is it that there's a fixed reference pressure array, and the "perturbation pressure" is the deviation from that array at each point/time?

>So it doesn't cleanly fit in with the current options supported ('sigma' and 'pressure'). This possibly puts a little more incentive on resolving #80, and having a bit more of a general framework for handling vertical coordinates (outside the scope of this PR).

Yes, I agree.  This is another instance where a user can end up "stuck" with no way to use aospy.  We definitely need to introduce a means by which to support arbitrary vertical coordinates.

spencerkclark:
> Interesting. So is it that there's a fixed reference pressure array, and the "perturbation pressure" is the deviation from that array at each point/time?

That's my understanding (see the description of the "total pressure in mb" on page 5-131 of the [WRF User's Guide](http://www2.mmm.ucar.edu/wrf/users/docs/user_guide_V3.9/ARWUsersGuideV3.9.pdf)); here's an example of the output from WRF:

```
In [3]: ds.P
Out[3]:
<xarray.DataArray 'P' (Time: 360, bottom_top: 49, south_north: 95, west_east: 95)>
array([[[[-40.028294, ..., -42.418442],
         ...,
         [-42.895279, ..., -42.370758]],

        ...,
        [[ -0.854176, ...,  -0.81953 ],
         ...,
         [ -0.921231, ...,  -0.899997]]],


       ...,
       [[[-33.781727, ..., -34.401615],
         ...,
         [-34.336052, ..., -31.916101]],

        ...,
        [[ -0.907447, ...,  -0.881743],
         ...,
         [ -0.887331, ...,  -0.857156]]]], dtype=float32)
Coordinates:
    XLAT     (Time, south_north, west_east) float32 36.605 36.605 36.605 ...
    XLONG    (Time, south_north, west_east) float32 -97.485 -97.485 -97.485 ...
Dimensions without coordinates: Time, bottom_top, south_north, west_east
Attributes:
    FieldType:    104
    MemoryOrder:  XYZ
    description:  perturbation pressure
    units:        Pa
    stagger:

In [4]: ds.PB
Out[4]:
<xarray.DataArray 'PB' (Time: 360, bottom_top: 49, south_north: 95, west_east: 95)>
array([[[[ 99630.460938, ...,  99630.460938],
         ...,
         [ 99630.460938, ...,  99630.460938]],

        ...,
        [[  4280.646484, ...,   4280.646484],
         ...,
         [  4280.646484, ...,   4280.646484]]],


       ...,
       [[[ 99630.460938, ...,  99630.460938],
         ...,
         [ 99630.460938, ...,  99630.460938]],

        ...,
        [[  4280.646484, ...,   4280.646484],
         ...,
         [  4280.646484, ...,   4280.646484]]]], dtype=float32)
Coordinates:
    XLAT     (Time, south_north, west_east) float32 36.605 36.605 36.605 ...
    XLONG    (Time, south_north, west_east) float32 -97.485 -97.485 -97.485 ...
Dimensions without coordinates: Time, bottom_top, south_north, west_east
Attributes:
    FieldType:    104
    MemoryOrder:  XYZ
    description:  BASE STATE PRESSURE
    units:        Pa
    stagger:
```

For this particular configuration of WRF I think the base state pressure just varies in 1D (in the vertical), even though this 1D array is broadcast to all columns and times in the model diagnostic.

spencerahill:
Thanks for that info.  I've opened #184 to track this.

spencerahill:
Also are those weird dimension names ('bottom_top', 'south_north', etc.) going to bite us?

spencerkclark:
@spencerahill yes, but I think those *may* be best addressed in a preprocessing function.  Let me give a little more background.

This particular configuration of WRF is a bit tricky to deal with, because it is a limited area model (see [Wang and Sobel, 2011](http://www.ldeo.columbia.edu/~sobel/Papers/wang_sobel_11.pdf)) with a rectangular domain.  All the gridboxes have the same area, and (somewhat pathologically) XLONG and XLAT have the same value at all grid points (even though they are defined on the dimensions Time, south_north, and west_east).  

```
In [6]: ds.XLAT.min()
Out[6]:
<xarray.DataArray 'XLAT' ()>
array(36.60499954223633)

In [7]: ds.XLAT.max()
Out[7]:
<xarray.DataArray 'XLAT' ()>
array(36.60499954223633)

In [8]: ds.XLONG.min()
Out[8]:
<xarray.DataArray 'XLONG' ()>
array(-97.48500061035156)

In [9]: ds.XLONG.max()
Out[9]:
<xarray.DataArray 'XLONG' ()>
array(-97.48500061035156)
```

There are a few things that will require some preprocessing before they are compatible with aospy.  

1. A CF-compliant units attribute to the time variable must be added
2. A dummy (i.e. say all ones) surface area coordinate must be added (there is no real way of computing it from the provided data)

In principle, one could probably take a domain average of a quantity out of the box with aospy without modifying the coordinates at all.  Essentially one could simply define a region that included the single (lat, lon) pair represented by XLONG and XLAT, say:

```python
globe = Region(
    name='globe',
    description='Entire globe',
    lat_bounds=(-90, 90),
    lon_bounds=(0, 360),
    do_land_mask=False
)
```

and things might work.  

But this would not enable finer specification of parts of the domain (say if one used non-uniform surface boundary conditions).  To enable this, an option would be to define new lat and lon coordinates that were 1D (rather than unnecessarily 3D like XLONG and XLAT), and tied directly to the dimensions (i.e. rename "north_south" "lat" and rename "west_east" "lon").  One could assign values to these dummy coordinates as the linspace between 0 and 1 along each dimension and then define regions from that perspective (normalizing would allow for the use of different horizontal domain sizes for instance, with regions taking up the same fraction of the domain).  This of course would need to be done as an optional third step in the preprocessing procedure.

A bit delicate, but I think doable?

spencerahill:
Thanks for digging into this.  A lot to consider.  I have more questions than ideas at this point.

>it is a limited area model

This is a big one.  We have been thinking almost exclusively about global domains with lat-lon grids.  Certainly we want to extend to non-global-sphere and/or non-lat-lon domains, but it might not be without some headaches.

>A dummy (i.e. say all ones) surface area coordinate must be added (there is no real way of computing it from the provided data)

Can the x and y extents of each gridbox be inferred?  E.g. something analogous to `lat_bounds` and `lon_bounds`?  Are the x and y points equally spaced?

>one could probably take a domain average of a quantity out of the box with aospy without modifying the coordinates at all. Essentially one could simply define a region that included the single (lat, lon) pair represented by XLONG and XLAT

Good point.

>But this would not enable finer specification of parts of the domain (say if one used non-uniform surface boundary conditions).

Even better point.

>define new lat and lon coordinates that were 1D (rather than unnecessarily 3D like XLONG and XLAT), and tied directly to the dimensions (i.e. rename "north_south" "lat" and rename "west_east" "lon")

So, in essence, for convenience we'd be using 'lat' and 'lon' within aospy as shorthand for the y and x coordinates more generally.  Right?

>One could assign values to these dummy coordinates as the linspace between 0 and 1 along each dimension and then define regions from that perspective (normalizing would allow for the use of different horizontal domain sizes for instance, with regions taking up the same fraction of the domain)

That could work.  Another idea I had in contemplating how to represent the "global" region (i.e. the entire model domain) in a non lat-lon framework would be for Region to accept slices, e.g. `lat_bounds=slice(0, None)` would grab that entire dimension.  We could also accept e.g. the string 'all' as shorthand.  I'm not sure how useful those would actually be though...

spencerkclark:
> Can the x and y extents of each gridbox be inferred? E.g. something analogous to lat_bounds and lon_bounds? Are the x and y points equally spaced?

Unfortunately I don't think there are the equivalent of lat and lon bounds in the model output; that being said, I *think* all the grid cells have the same area (so simply assigning equal weights, a.k.a. surface areas, is OK in this use-case).

> So, in essence, for convenience we'd be using 'lat' and 'lon' within aospy as shorthand for the y and x coordinates more generally. Right?

Yes, in an ideal world I agree we would probably want to rename the dimensions to reflect the difference in geometry, but I think this is at least the quickest way to get up and running.

> That could work. Another idea I had in contemplating how to represent the "global" region (i.e. the entire model domain) in a non lat-lon framework would be for Region to accept slices, e.g. lat_bounds=slice(0, None) would grab that entire dimension.

Can you mask with a slice?  I suppose you could slice along a dimension and reindex back to the full coordinate with no fill method and call `np.isfinite` on the result, but we'd need to think carefully how we would handle multi-dimensional latitude and/or longitude coordinates in that setup (we might have to do some stacking and unstacking).

spencerahill:
>I think all the grid cells have the same area (so simply assigning equal weights, a.k.a. surface areas, is OK in this use-case

There's some relevant logic already in [Model._grid_sfc_area](https://github.com/spencerahill/aospy/blob/develop/aospy/model.py#L75) and the methods it calls.  If no bound arrays are found, it constructs them based on the spacing of the array values.  However, I also see that [we have the spherical geometry baked in](https://github.com/spencerahill/aospy/blob/develop/aospy/model.py#L89).  So some refactoring and additional logic there will be necessary.

Ultimately we need to somehow specify or determine the geometry.  How to do so?  A new `geometry` attribute for the core object hierarchy, with default 'sphere'?  Perhaps eventually we could try inferring it from the coordinate names, but that's hard, because I've seen some obs data that uses 'x' and 'y' for lon and lat, respectively.

>Can you mask with a slice?

Ya, let's forget that for now.  Pretty sure it doesn't actually make a ton of sense.  Although I do still like the 'all' keyword.

Another new thought: your fractional coordinate idea could be nice for Regions.  E.g. represent the global region as (0, 1) and (0, 1) x- and y- ranges respectively, the northern hemisphere as (0, 1) and (0.5, 1), etc.  This could be agnostic to the coordinate values, unlike the current formulation that assumes -90 to 90 in lat and 0 to 360 in lon.

PR #186, opened 2017-05-30
WIP add support for time-dependent land masks

### LINKED ISSUE #181: Support time-dependent land masks
I have just come across this use case for the first time: the land mask is defined at each timestep, i.e. (time, lat, lon).  We currently assume that the land mask is defined only in (lat, lon).  The time-dependent case is more general and so we should support it.

I don't foresee this being massively difficult; it just throws off some of our existing logic a bit.  Will put together a PR hopefully relatively soon.

spencerahill:
Closes #181 

Originally I thought that the land mask of the data I was working with was defined in time but with identical values across timesteps, but the land mask does actually vary slightly with time (apparently based on sea level, which is cool).  Anyways, before I realized that, I made this `_maybe_truncate_dim` logic.  However, since that's no longer a use case, I'm inclined to drop it.

Also, my new logic in _avg_by_year feels clunky.  @spencerkclark any way to clean this up?  The problem is that when land_mask is a coordinate with time dependence, it gets dropped by the groupby-apply-combine step.  So my solution is to promote it to a variable, perform the groupby arithmetic, and then convert back to a DataArray with it as a coordinate.

Apologies for a few PEP8 fixes that I snuck in here.  Also need tests -- I've only checked that this is working to the extent that I could submit a calculation with a time-dependent mask that didn't crash.

### DIFF
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -41,6 +41,22 @@ ps = Var(
 )
 
 
+def _maybe_truncate_dim(arr, dim=internal_names.TIME_STR):
+    """Drop a given dimension if the data is constant in that dimension."""
+    try:
+        [xr.testing.assert_equal(arr.isel(**{dim: 0}),
+                                 arr.isel(**{dim: i}))
+         for i in range(len(arr[dim]))]
+    except AssertionError:
+        logging.debug("Unable to drop dimension '{0}' of array '{1}', due to "
+                      "it not being invariant.".format(dim, arr))
+        return arr
+    else:
+        logging.info("Dropped dimension '{0}' of array '{1}', due to "
+                     "it being invariant.".format(dim, arr))
+        return arr.drop(dim)
+
+
 class CalcInterface(object):
     """Interface to the Calc class."""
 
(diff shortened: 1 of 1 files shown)


spencerkclark:
@spencerahill I'm fine either way including `_maybe_truncate_dim`; it might actually be useful for other things later on (e.g. I can think of some use-cases in WRF where some coordinates are unnecessarily time-dependent).

For simplifying logic in `_avg_by_year` two options come to mind (I think you're right that there is no way around promoting the land mask to a variable):
1. Given xarray's automatic broadcasting capabilities, you could make the assumption that all land masks were time-dependent (and always promote them to being variables to get an annual time series); this would cut down on the if statements, but would unnecessarily broadcast non-time-dependent landmasks to be time dependent (not really ideal).
2. Encapsulate more logic into potentially reusable functions.  For instance, it might be useful to create a general function that promoted certain coordinates to variables when applying a certain function, and then set them back to coordinates when finished.  Maybe something like:

```python
def apply_with_coords_as_variables(f, coords, da, *args, **kwargs):
    """Apply a function to a DataArray and a subset of its coordinates
    
    Parameters
    ----------
    f : function
        Function (must take an xr.Dataset as first argument and
        return an xr.Dataset)
    coords : str or list
        Coordinates to promote to variables
    da : xr.DataArray
        DataArray to apply function to
    *args
        Arguments to pass to f
    **kwargs
        Keyword arguments to pass to f
        
    Returns
    -------
    xr.DataArray
    """
    name = da.name
    ds = da.reset_coords(coords)
    ds = f(ds, *args, **kwargs)
    return ds.set_coords(coords)[name]
```

Then within `_avg_by_year` you could have logic that would look like:
```python
if mask_is_time_defined:
    return apply_with_coords_as_variables(
        yr_avg, internal_names.LAND_MASK_STR, arr, dt, yr_str)
else:
    return yr_avg(arr, dt, yr_str)
```
assuming you defined a function `yr_avg` to look like:
```python
def yr_avg(ds, dt, yr_str):
    return ((ds*dt).groupby(yr_str).sum(internal_names.TIME_STR) /
```

I might be inclined to go with something like option (2).  Though it might be more code than you have now, I think it makes `_avg_by_year` a little more readable, and might make things easier to test.  What are your thoughts?

spencerahill:
>it might actually be useful for other things later on (e.g. I can think of some use-cases in WRF where some coordinates are unnecessarily time-dependent).

I agree.  But to keep the scope limited (I'm pretty time crunched for the next several weeks), I think I'll leave it out for now.  We can come back to the last commit as a starting point if/when we want to implement it.

>Encapsulate more logic into potentially reusable functions. For instance, it might be useful to create a general function that promoted certain coordinates to variables when applying a certain function, and then set them back to coordinates when finished.

I agree, this seems like the way to go.  Thanks a lot for all the code snippets, they provide a great starting point.  Will pursue this.

Issue #98, opened 2016-10-24
Long term solution for limited permissible dates due to np.datetime64 nanosecond resolution


spencerahill:
E.g. #96.  This maybe is a wait-and-see for upstream solutions (either at xarray, pandas, or numpy levels), but wanted to have an Issue open where we can track this/discuss alternatives.

Our current workarounds limit us to <585 years of data, which for many folks in the community would be wholly insufficient.

spencerkclark:
[It seems that this will likely not change anytime soon in pandas](https://github.com/pandas-dev/pandas/issues/7307).  In the meantime, [their suggested workaround in the docs](http://pandas.pydata.org/pandas-docs/stable/timeseries.html#timeseries-oob]) is to use `pd.Period` for dates outside the valid range.  On the topic of alternatives -- here is a summary of the existing options (as far as I can tell, none work for all our requirements).

## Xarray's Workaround

As we know, xarray doesn't officially use the pandas suggested workaround; instead it uses `netCDF4.datetime` objects in this situation.  The xarray solution is a compromise, because one cannot do time-related groupby operations when using `netCDF4.datetime` objects in the time variable.  E.g.

``` python
In [1]: import xarray as xr

In [2]: da = xr.DataArray([15, 45, 74], coords=[[15, 45, 74]], dims=['time'], name='a')

In [3]: da['time'].attrs['units'] = 'days since 0001-01-01'

In [4]: da = da.to_dataset()

In [5]: da = xr.decode_cf(da)
//anaconda/lib/python2.7/site-packages/xarray/conventions.py:377: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  result = decode_cf_datetime(example_value, units, calendar)

In [6]: da.groupby('time.month').mean('time')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-b8680c841459> in <module>()
----> 1 da.groupby('time.month').mean('time')

//anaconda/lib/python2.7/site-packages/xarray/core/common.pyc in groupby(self, group, squeeze)
    342         """
    343         if isinstance(group, basestring):
--> 344             group = self[group]
    345         return self.groupby_cls(self, group, squeeze=squeeze)
    346

//anaconda/lib/python2.7/site-packages/xarray/core/dataset.pyc in __getitem__(self, key)
    518
    519         if hashable(key):
[...]

PR #189, opened 2017-06-08
Fix workaround for years less than 1678 for non-0001 reference dates

### LINKED ISSUE #188: BUG: Workaround for datetimes with years less than 1678 not valid for reference date not set to 0001-01-01
This came up in helping set up aospy for someone looking at an AM2 run which was run for years 781 to 800.  The reference date in the units attribute of the time variable in the output files was:
```
double time(time) ;
		time:long_name = "time" ;
		time:units = "days since 0761-01-01 00:00:00" ;
		time:cartesian_axis = "T" ;
		time:calendar_type = "NOLEAP" ;
		time:calendar = "NOLEAP" ;
		time:bounds = "time_bounds" ;
```
January 1st, 0761.  

The current logic in `utils.times.numpy_datetime_workaround_encode_cf` looks like:

```python
def numpy_datetime_workaround_encode_cf(ds):
    time = ds[internal_names.TIME_STR]
    units = time.attrs['units']
    units_yr = units.split(' since ')[1].split('-')[0]
    min_yr_decoded = xr.decode_cf(time.to_dataset(name='dummy'))
    min_date = min_yr_decoded[internal_names.TIME_STR].values[0]
    if isinstance(min_date, np.datetime64):
        return ds, pd.Timestamp(min_date).year
    else:
        min_yr = min_date.year
        new_units_yr = pd.Timestamp.min.year + 2 - min_yr
        new_units = units.replace(units_yr, str(new_units_yr))

        for VAR_STR in internal_names.TIME_VAR_STRS:
            if VAR_STR in ds:
                var = ds[VAR_STR]
                var.attrs['units'] = new_units
        return ds, min_yr
```

In the case of these files, `min_yr` was 781; therefore `new_units_yr` was set to 1679 - 781 = 898.  This made the new reference date in the files `0898-01-01 00:00:00`.  But, since the raw time values (originally encoded as `"days since 0761-01-01 00:00:00"`) had values
```
time = 7315.5, 7345, 7374.5, 7405, 7435.5, 7466, 7496.5, 7527.5, 7558,
    7588.5, 7619, 7649.5, 7680.5 ...
}
```
the newly decoded times (using the new reference date), still were not in the Timestamp-valid range.  E.g. 7315.5 is only roughly 20 years, so the first value would be decoded as in year 918 rather than in 1678.

I *think* a solution to this would be to change the line `new_units_yr = pd.Timestamp.min.year + 2 - min_yr` to `new_units_yr = pd.Timestamp.min.year + 1 - min_yr + int(units_yr)` in the above method, but I haven't thought through it carefully yet.

spencerkclark:
Closes #188

@spencerahill I can confirm the tests added here failed prior to making this change.


aospy/utils/times.py
```diff
@@ -198,13 +198,14 @@ def numpy_datetime_workaround_encode_cf(ds):
     time = ds[internal_names.TIME_STR]
     units = time.attrs['units']
     units_yr = units.split(' since ')[1].split('-')[0]
+    ref_yr = int(units_yr)
     min_yr_decoded = xr.decode_cf(time.to_dataset(name='dummy'))
     min_date = min_yr_decoded[internal_names.TIME_STR].values[0]
     if isinstance(min_date, np.datetime64):
         return ds, pd.Timestamp(min_date).year
     else:
         min_yr = min_date.year
```
spencerahill: After defining `min_yr`, let's define `offset = int(units_yr) - min_yr + 1` rather than `ref_yr` and include a comment above it describing it, e.g. `# Offset such that decoded dates start in the earliest valid whole year`.  Then the line below becomes `new_units_yr = pd.Timestamp.min.year + offset`.

spencerahill:
Thanks for the quick turnaround on this.  Can you also add a test with non-January 1st month and day?

Potential edge cases that come to mind: dates in 1678, both in and out of the valid range; same for dates in the last valid year.

spencerkclark:
>Potential edge cases that come to mind: dates in 1678, both in and out of the valid range; same for dates in the last valid year.

Yeah a minimum date with the same year as the Timestamp minimum date would cause problems (I think it's actually 1677).  In its current form, the logic that we use to transform dates from user input to the workaround versions (`numpy_datetime_range_workaround`) would also fail for dates larger than the Timestamp maximum.  Do you think we should address that or wait until someone has a use case for it?

aospy/test/test_utils_times.py
```diff
@@ -144,57 +144,60 @@ def test_numpy_datetime_range_workaround(self):
         )
 
     def test_numpy_datetime_workaround_encode_cf(self):
+        def create_test_data(
+                days, ref_units, expected_units, expected_date):
```
spencerahill: `expected_date` isn't used.

Nit: non-standard to start the func args on the next line.  I would just break after `ref_units` to stay within 80 chars.

spencerahill:
>Do you think we should address that or wait until someone has a use case for it?

Your call based on how easy the fix would be and your time available.  Although if we're saying we've identified a bug we should at least warn/document it somehow.

spencerkclark:
@spencerahill I think I fixed it (see the commit message for a more detailed explanation), but I think we should think carefully about:

- When this workaround breaks down (so we can raise an informative exception)
- Whether what we have implemented currently is the absolute best we can do (as far as a workaround goes)
- Whether this has adequate tests

before merging.  Hopefully I'll have some time over the next few days, but I obviously appreciate your feedback as well.

spencerahill:
>Originally this method used the user-input subset bounds.  This
would have the potential of being an issue if the dataset included
data from times before 1678, say 1600, but the user only wanted to subset
data between say 1679 and 1750.  In that case internally, dates
would be offset to start in 1678, but the subset dates would not be
offset.

Good catch!

>Whether what we have implemented currently is the absolute best we can do (as far as a workaround goes)

Let's also keep in mind that all of this is temporary; it will all go away once #98 / https://github.com/pydata/xarray/issues/1084 are addressed.  As such, I think it's reasonable to fix bugs in actual use cases as they come up, but potentially just warn/raise when we recognize edge cases that aren't actually affecting users so far, e.g. this edge case of subsetting within the valid range from input files extending outside the valid range.

That's my take.  What's the latest with #98 and https://github.com/pydata/xarray/issues/1084, and based on that do you think my take seems reasonable?

I'll give a more careful code review once we converge on these bigger picture aspects.

spencerkclark:
> Let's also keep in mind that all of this is temporary; it will all go away once #98 / pydata/xarray#1084 are addressed.

Right.  pydata/xarray#1252 is the place to look for progress on that.  The updates I pushed about a month ago now are a first pass at resetting the logic in xarray to use a NetCDFTimeIndex whenever a non-standard calendar is used or dates with standard calendar are outside the Timestamp-valid range.  I think the gist of what we set out to accomplish is there (e.g. partial-string indexing, saving to netCDF, and field accessors), but it needs a good deal of cleaning up (and I could probably use some guidance on how best to do that). 

 > As such, I think it's reasonable to fix bugs in actual use cases as they come up, but potentially just warn/raise when we recognize edge cases that aren't actually affecting users so far, e.g. this edge case of subsetting within the valid range from input files extending outside the valid range.

For this particular edge case it may actually be just as easy to just fix it as it would be to warn/raise upon encountering it :).  To warn/raise I think we would still need to check both `min_yr` and `max_yr` in `numpy_datetime_range_workaround` (checks which I've introduced as part of the fix).  Bigger picture though, I agree, for future edge cases we may think of (that aren't impacting users) warning/raising is probably plenty.

spencerahill:
>it needs a good deal of cleaning up (and I could probably use some guidance on how best to do that)

Indeed.  I just pinged that thread to hopefully rekindle the discussion 😄 

>For this particular edge case it may actually be just as easy to just fix it as it would be to warn/raise upon encountering it :).

That's totally fine...of course all else equal a fix is better than a warning!  I'll review your code now.

aospy/utils/times.py
```diff
@@ -165,7 +167,9 @@ def numpy_datetime_range_workaround(date, min_year):
         permissible dates, otherwise a datetime.datetime object with the year
         offset to the earliest allowable year.
     """
-    if date < pd.Timestamp.min:
+    min_yr_in_range = pd.Timestamp.min.year < min_year < pd.Timestamp.max.year
+    max_yr_in_range = pd.Timestamp.min.year < max_year < pd.Timestamp.max.year
+    if not (min_yr_in_range and max_yr_in_range):
```
spencerahill, line 172: Just confirming I'm understanding the situation correctly: if any part of the loaded data is outside of the valid range, the decoding will be unable to use the standard numpy datetimes.  In that case, we apply our offset so that it "starts" near the beginning of the valid date range.  

Importantly, the loaded data includes the whole range of data loaded onto disk, even if the date range we're interested in is a subset of that, and even if that subset is itself entirely within the valid range.

If that's correct then I think your logic is correct.

aospy/test/test_data_loader.py
```diff
@@ -442,6 +443,26 @@ def test_load_variable(self):
         expected = xr.open_dataset(filepath)['condensation_rain']
         np.testing.assert_array_equal(result.values, expected.values)
 
+    def test_load_variable_non_0001_refdate(self):
+        def preprocess(ds, **kwargs):
+            ds['time'] = ds['time'] - 1095.
```
spencerahill: Can you add a comment explaining the significance/reasoning for the 1095 value?  And/or define it as a constant.

aospy/test/test_utils_times.py
```diff
@@ -130,49 +130,118 @@ def test_datetime_or_default(self):
...
+            ds[TIME_STR].attrs['calendar'] = 'noleap'
+            actual, min_yr, max_yr = numpy_datetime_workaround_encode_cf(ds)
+
+            time_desired = xr.DataArray([days, days + 1095.], dims=[TIME_STR])
+            desired = xr.Dataset(coords={TIME_STR: time_desired})
+            desired[TIME_STR].attrs['units'] = expected_units
+            desired[TIME_STR].attrs['calendar'] = 'noleap'
+            return actual, min_yr, max_yr, desired
+
+        # 255169 days from 0001-01-01 corresponds to date 700-02-04.
```
spencerahill, line 181: Something like this line is what I have in mind in the above comment re: the 1095 value

spencerahill (commented):
A few minor things, otherwise this looks good to me.  Thanks!

aospy/utils/times.py
spencerkclark: That's exactly right

spencerkclark:
@spencerahill thanks for having a look and for pinging the NetCDFTimeIndex thread; I think I took care of what you were looking for here.

Let me know if you have any more comments!

spencerahill:
Awesome, thanks @spencerkclark!  Nice work as usual.

PR #192, opened 2017-06-20
Decode missing values in DataLoader and make regions respect existing missing values

### LINKED ISSUE #190: Masked input data problems, both gridpoint-by-gridpoint and for regional reductions
(summarizing offline convo among myself, @spencerkclark , and a new aospy user(!))

Data interpolated to pressure levels is generally masked at those points/times where that box is below ground.  Within the netCDF files, this is often flagged using an extremely large magnitude (positive or negative) number.  Currently aospy is not honoring that mask, and these large values are being treated as OK, which obviously is not good.

The likely culprit is `mask_and_scale=False` in our [reading in of input data files](https://github.com/spencerahill/aospy/blob/develop/aospy/data_loader.py#L161-L163).  But we still need to fully test it.

A related issue is that the [regional averages](https://github.com/spencerahill/aospy/blob/develop/aospy/region.py#L164-L171) don't appear to honor these masks either.  We suspect that replacing the current `weights=...` logic with `weights = _sum_over_lat_lon(sfc_area*land_mask).where(np.isfinite(data))` will do the trick, but once again we haven't fully tested it yet.

I don't foresee a ton of difficulty finishing resolving these, but did want to track them here.

spencerkclark:
Closes #190


aospy/region.py
```diff
@@ -166,7 +168,8 @@ def ts(self, data):
         data_masked = self.mask_var(data)
         sfc_area = data.sfc_area
         land_mask = _get_land_mask(data, self.do_land_mask)
-        weights = _sum_over_lat_lon((self.mask_var(sfc_area)*land_mask))
+        weights = _sum_over_lat_lon(
```
spencerahill: Could you add a comment that motivates the `np.isfinite` inclusion?

aospy/test/test_region.py
```diff
@@ -0,0 +1,93 @@
...
+import xarray as xr
+
+from aospy import Region
+from aospy.region import _get_land_mask
+from aospy.internal_names import (LAT_STR, LON_STR,
+                                  SFC_AREA_STR, LAND_MASK_STR)
+
+
+@pytest.fixture()
+def data():
```
spencerahill: Let's rename this to something more descriptive, e.g. `data_for_reg_calcs`

aospy/test/test_region.py
```diff
@@ -0,0 +1,93 @@
...
+    return Region(
+        name='test',
+        description='Test region with no land mask',
+        lat_bounds=(0., 90.),
+        lon_bounds=(0., 5.),
+        do_land_mask=True
+    )
+
+
+def test_mask_var(data, region_no_land_mask):
```
spencerahill: Can we add an analogous test for `region_land_mask` for completeness

spencerahill: Or maybe incorporate it into this test via pytest's parameterize functionality

aospy/test/test_region.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_get_land_mask_with_land_mask(data, region_land_mask):
+    result = _get_land_mask(data, region_land_mask.do_land_mask)
+    expected = data[LAND_MASK_STR]
+    xr.testing.assert_identical(result, expected)
+
+
+def test_ts_no_land_mask(data, region_no_land_mask):
+    result = region_no_land_mask.ts(data)
+    expected = xr.DataArray((0.25 * 4. + 0.5 * 3.) / (0.25 + 0.5))
```
spencerahill: Can you add a note explaining where these values come from?  And/or better yet pull them directly from the test data.

spencerahill (commented):
Thanks @spencerkclark for taking the time to do this, especially the tests.  A few minor things, mostly for clarification, but otherwise looks good.

spencerkclark:
@spencerahill thanks for the review comments; I went ahead and fixed the `mask_and_scale` issue as well (and added a test).  This should fully take care of #190 now.  Let me know if you have any more comments.

aospy/test/test_data_loader.py
```diff
@@ -492,6 +492,24 @@ def preprocess(ds, **kwargs):
         expected = xr.open_dataset(filepath)['condensation_rain']
         np.testing.assert_allclose(result.values, expected.values)
 
+    def test_load_variable_mask_and_scale(self):
+        def preprocess(ds, **kwargs):
```
spencerahill: Name this `convert_all_to_missing_val` in place of the comment.

aospy/test/test_data_loader.py
```diff
@@ -492,6 +492,24 @@ def preprocess(ds, **kwargs):
...
+            return ds
+
+        self.data_loader.preprocess_func = preprocess
+
+        result = self.data_loader.load_variable(
+            condensation_rain, datetime(5, 1, 1),
+            datetime(5, 12, 31),
+            intvl_in='monthly')
+
+        result = np.isfinite(result).sum().item()
```
spencerahill: Either rename this or the above `result` to a different name.  E.g. for this one `result_integral`.

spencerahill (commented):
Thanks @spencerkclark, nice work as usual.  Two final nits, then please go ahead and merge this. 👍

spencerkclark:
Thanks @spencerahill, merging now.

PR #195, opened 2017-07-27
Dask async warning

spencerahill:
As of dask v0.15, their ``async`` module that we reference [has been renamed](https://github.com/dask/dask/pull/2318) to ``local``.  This PR checks the dask version so that we import from the correct place, thereby avoiding a deprecation message about it.

@spencerkclark I don't think this warrants tests, but let me know if you think otherwise.

(Sorry for the accidental incoherent commit message)


aospy/model.py
```diff
@@ -10,8 +10,14 @@
 from . import utils
 
 # Dask must use its serial scheduler if computations are to be performed
-# in parallel using multiprocess
-dask.set_options(get=dask.async.get_sync)
```
spencerkclark: @spencerahill good catch!  It's possible this line may not even be needed anymore now that we use `dask.bag` for parallel computations.  Could you see if the test suite passes with it removed?  (In other words we may not need the workaround in this PR).

See [this line](https://github.com/spencerahill/aospy/pull/172/files#diff-c1bc79d9c0310c1c1c15a5943463101cL14) in #172, where this code was already removed from `data_loader.py`.  Sorry if I missed this there.

spencerahill: @spencerkclark you're right!  Even better.  Thanks for catching that.

spencerkclark (approved):
Looks good to me!  Thanks for taking care of this.

Issue #75, opened 2016-05-21
Prevent Empty Header error when two processes try to write to same .tar file


spencerahill:
Sometimes I will have aospy computing things at the same time on multiple analysis nodes (or the same node but from different terminal sessions).  This often produces low-level errors when writing the data to the /archive data.tar files, e.g. "Empty header file".  I think it's simply a result of one session trying to access/modify the tar file at the same time as another one.

There should be some way of implementing a lock to prevent this from happening.  How to test seems more tricky.

spencerahill:
This is becoming more of an issue now that we have Calcs fired off in parallel.  I just tried submitting a main script that specified 4 variables computed in 4 different runs in a single model, and 3/3 times I tried it I got either the empty header file error or a string of "tar: Deleting non-header from archive" messages interspersed with a few "tar: Exiting with failure status due to previous errors", ultimately ending in "tarfile.ReadError: bad checksum"

@spencerkclark has this ever happened to you?  Googling the error messages didn't lead to anything immediately helpful.

dboeckenhoff:
I have the same problem with a similar usecase. Have you encountered any solution yet?

spencerahill:
Unfortunately I have not, in part because I haven't spent any time on this since the last post.  I just did a bit of googling, and [this SO post](https://stackoverflow.com/questions/489861/locking-a-file-in-python) seems to be a useful starting point.  This is assuming that I'm correct that we need a "lock" of some kind -- frankly I'm pretty ignorant here.  

Is your usecase when using aospy or no?  If not, you might look into this [FileLock](https://github.com/dmfrey/FileLock) package listed in the above link.  For aospy, we'll have to think about the design.

@spencerkclark, any ideas on this?

spencerkclark:
I think this is the root cause of some intermittent test failures in the CI https://github.com/spencerahill/aospy/pull/180#issuecomment-300133662 , so it would be good to solve this once and for all.

@spencerahill I too have not thought much about how to fix this.  That FileLock package seems like it could be a nice simple solution, but I agree we'll have to think about it more for aospy.

spencerahill:
Or can we use something in the standard library, e.g. something within asyncio?  Or something in dask?

This feels hackish, but one could put a try/except clause within a loop, with a pause between each attempt and a maximum number of attempts:

```python
max_attempts = 10
for i in range(max_attempts):
    try:
        # <insert logic for writing to .tar>
    except HeaderError as e:  # or whatever the exception class is
[...]

PR #196, opened 2017-08-31
Retain original data's mask when yearly averaging

### LINKED ISSUE #193: Bug: not accounting for masked values in computing weights for temporal averages
Similar to #190 but for temporal rather than spatial averages.  Paraphrasing @spencerkclark from offline:

We also need to make sure the time weights are masked where data points are invalid whenever we take time averages.  A quick one-line change (in red) in calc.py seemed to fix things:

```python
def _avg_by_year(self, arr, dt):
    """Average a sub-yearly time-series over each year."""
    utils.times.assert_matching_time_coord(arr, dt)
    yr_str = internal_names.TIME_STR + '.year'
    dt = dt.where(np.isfinite(arr))  # this is the one-line change
    return ((arr*dt).groupby(yr_str).sum(internal_names.TIME_STR) /
```

but we'll need to think things through and add tests to make sure that this solves the problem!

spencerahill:
Closes #193.

I still need a test for the actual masking issue; the current test checks un-masked data.

Re-factored a bit in the process, namely moved the function from `Calc` into `utils.times`.


spencerahill:
I'm not sure what to make of [these failures](https://travis-ci.org/spencerahill/aospy/jobs/270270575) in the Travis 2.7-min environment.  Not related to the current PR but also not the same as the ongoing issues c.f. #75 

The [AppVeyor failures](https://ci.appveyor.com/project/spencerahill/aospy/build/1.0.285/job/g6mp5r191q698hpr#L404) are for python 3.5 and are similar-but-different from either those above or the EmptyHeader of #75.

@spencerkclark any ideas on these?  Plus a quick review of the PR when you get a chance please :)

aospy/test/test_utils_times.py
```diff
@@ -485,5 +486,51 @@ def test_sel_time(self):
...
+    actual = yearly_average(arr, dt)
+
+    yr2000 = (arr[0]*dt[0] + arr[1]*dt[1]) / (dt[0] + dt[1])
+    yr2001 = ((arr[2]*dt[2] + arr[3]*dt[3] + arr[4]*dt[4]) /
+              (dt[2] + dt[3] + dt[4]))
+    yr2004 = arr[-1]
+    yrs_coord = [2000, 2001, 2004]
+    yr_avgs = np.array([yr2000, yr2001, yr2004])
+    desired = xr.DataArray(yr_avgs, dims=['year'], coords={'year': yrs_coord})
+    assert np.allclose(actual.values, desired.values)
```
spencerkclark: Both here and in the masked test, could we use `xr.testing.assert_identical(actual, desired)`?  I think this would check the array values and year coordinate all in one line.

docs/whats-new.rst
```diff
@@ -72,6 +72,9 @@ Bug Fixes
   datapoints were invalid instead of just masking points outside the desired
   region (fixes :issue:`190` via :pull:`192`).  By
   `Spencer Clark <https://github.com/spencerkclark>`_.
+- Retain original input data's mask during gridpoint-by-gridpoint
+  tempral averages (fixes :issue:`193` via :pull:`196`).  By `Spencer
```
spencerkclark: Typo: "tempral" -> "temporal"

spencerkclark (commented):
Thanks for taking care of this @spencerahill!  Just a couple minor comments, but otherwise this looks good to me.

spencerkclark:
> I'm not sure what to make of these failures in the Travis 2.7-min environment. Not related to the current PR but also not the same as the ongoing issues c.f. #75

It seems they've encountered similar error messages in xarray and it's been suggested this is related to a recent update to `dask.distributed` (pydata/xarray#1540).  I think we can follow along there to see what the resolution might be.

> The AppVeyor failures are for python 3.5 and are similar-but-different from either those above or the EmptyHeader of #75.

I want to say these are still related to #75, but I'm not 100% sure.

aospy/test/test_utils_times.py
spencerahill: Good call; forgot about `xr.testing`.  I'll use `xr.testing.assert_allclose` though; in some cases I was getting failures because my by-hand averaging weren't bitwise identical with the xarray/numpy methods.

spencerahill:
Thanks @spencerkclark.  Will merge pending the tests (besides those that #197 will hopefully fix)

PR #199, opened 2017-09-01
Relax logic for comparing grid attributes between Run and Model objs

### LINKED ISSUE #187: Should all grid attributes potentially found in files loaded by a Run object be required to be accessible to Model objects?
Currently if a grid attribute that can be found in a Run object cannot be found in a Model object, aospy crashes, because `numpy.allclose` cannot compare an array and `None`.  (For example if `zsurf` is represented in a file loaded in by a Run object, but it does not exist in any of the Model grid files).
```
In [5]: import numpy as np

In [6]: np.allclose(np.array([1, 2, 3]), None)
<ipython-input-6-d64ad5a1c13a> in <module>()
----> 1 np.allclose(np.array([1, 2, 3]), None)

//anaconda/envs/xarray-dev/lib/python2.7/site-packages/numpy/core/numeric.pyc in allclose(a, b, rtol, atol, equal_nan)
   2370
   2371     """
-> 2372     res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
   2373     return bool(res)
   2374

//anaconda/envs/xarray-dev/lib/python2.7/site-packages/numpy/core/numeric.pyc in isclose(a, b, rtol, atol, equal_nan)
   2449
   2450     xfin = isfinite(x)
-> 2451     yfin = isfinite(y)
   2452     if all(xfin) and all(yfin):
   2453         return within_tol(x, y, atol, rtol)

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
I don't *think* this should be the desired behavior.  I don't see any reason why one should not use a coordinate that is defined in data read in using a Run object that doesn't exist in the Model object.

A simple fix would be to just add an additional condition to [line 278 in `aospy.calc._add_grid_attributes`](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L278) to make sure that `model_attr is not None`:
```python
def _add_grid_attributes(self, ds):
    """Add model grid attributes to a dataset"""
    for name_int, names_ext in self._grid_attrs.items():
        ds_coord_name = set(names_ext).intersection(set(ds.coords) |
                                                    set(ds.data_vars))
        model_attr = getattr(self.model, name_int, None)
        if ds_coord_name and (model_attr is not None):
...
```
@spencerahill what do you think?  This came up in adapting aospy to use WRF output.  I know historically we both have provided sample data files to each Model object we use (so this hasn't come up before), but now that we have enabled preprocessing functions (which do not get called on Model grid files, and I think this should remain this way) that often change or add coordinates, providing a sample data file to the Model object to fill out its grid attributes doesn't always make sense.

spencerkclark:
Fixes #187

The way I tested this was by adding a grid attribute (`'zsurf'`) of all zeros to the files used in some test `Run` objects:
- `00040101.precip_monthly.nc`
- `00050101.precip_monthly.nc`
- `00060101.precip_monthly.nc`

but leaving it off the files used in constructing their `Model` object:
- `00060101.sphum_monthly.nc`

Tests do not pass without the one line change I add in `calc.py`.


spencerahill:
Gotta love a one-line fix!  Any idea what's going on in the tests with the Notebook?  Otherwise this looks good to me.

spencerkclark:
@spencerahill I need to run now, but it seems that the notebook exposed a valid use-case that was problematic.  My commit fixes things, but I don't think it is the ideal solution.  I'll try and lay out the problem in more detail tomorrow so we can discuss the options.

spencerahill:
Sounds good.  No rush.

spencerkclark:
The notebook sets up some `aospy` objects in the following manner:
```python
...
from aospy.data_loader import DictDataLoader
file_map = {'monthly': os.path.join(rootdir, '000[4-6]0101.precip_monthly.nc')}
data_loader = DictDataLoader(file_map)

from aospy import Run
example_run = Run(
    name='example_run',
    description='Control simulation of the idealized moist model',
    data_loader=data_loader
)

from aospy import Model
example_model = Model(
    name='example_model',
    grid_file_paths=(
        os.path.join(rootdir, '00040101.precip_monthly.nc'),
        os.path.join(rootdir, 'im.landmask.nc')
    ),
    runs=[example_run]  # only one Run in our case, but could be more
)
...
```

As of this PR each `*precip_monthly.nc` file now contains a 2D `zsurf` variable:
```
In [1]: import xarray as xr

In [2]: single_file = xr.open_dataset('00040101.precip_monthly.nc', decode_cf=Fa
   ...: lse)

In [3]: single_file.zsurf
Out[3]:
<xarray.DataArray 'zsurf' (lat: 64, lon: 128)>
array([[ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       ...,
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.]], dtype=float32)
Coordinates:
  * lon      (lon) float64 0.0 2.812 5.625 8.438 11.25 14.06 16.88 19.69 ...
  * lat      (lat) float64 -87.86 -85.1 -82.31 -79.53 -76.74 -73.95 -71.16 ...
Attributes:
    _FillValue:     1e+20
    units:          m
    long_name:      Surface height
    time_avg_info:  average_T1,average_T2,average_DT
    cell_methods:   time: mean
```

The issue stems from how concatenation proceeds using `xr.open_mfdataset` in `data_loader.py`:

https://github.com/spencerahill/aospy/blob/ce7c78418811c9e0ffed7bc647f8df53c34e03ee/aospy/data_loader.py#L185-L190

When the three files are concatenated together using `xr.open_mfdataset`, a time dimension is added to `zsurf`:

```
In [4]: ds = xr.open_mfdataset('000[4-6]0101.precip_monthly.nc', decode_cf=False,
   ...: concat_dim='time').load()

In [5]: ds.zsurf
Out[5]:
<xarray.DataArray 'zsurf' (time: 36, lat: 64, lon: 128)>
array([[[ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.],
        ...,
        [ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.]],

       [[ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.],
        ...,
        [ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.]],

       ...,
       [[ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.],
        ...,
        [ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.]],

       [[ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.],
        ...,
        [ 0.,  0., ...,  0.,  0.],
        [ 0.,  0., ...,  0.,  0.]]], dtype=float32)
Coordinates:
  * lon      (lon) float64 0.0 2.812 5.625 8.438 11.25 14.06 16.88 19.69 ...
  * lat      (lat) float64 -87.86 -85.1 -82.31 -79.53 -76.74 -73.95 -71.16 ...
  * time     (time) float64 1.111e+03 1.139e+03 1.17e+03 1.2e+03 1.231e+03 ...
Attributes:
    _FillValue:     1e+20
    units:          m
    long_name:      Surface height
    time_avg_info:  average_T1,average_T2,average_DT
    cell_methods:   time: mean
```

Note that three files are specified in the `Run` object created in the notebook, and one `*precip_monthly.nc` file is specified in the `Model` object.  So the `zsurf` grid attribute in the `Run` object has a time dimension, but the `zsurf` grid attribute in the `Model` object does not.  When these reach the `Calc._add_grid_attributes` step, they are compared (the key line is 283):

https://github.com/spencerahill/aospy/blob/ce7c78418811c9e0ffed7bc647f8df53c34e03ee/aospy/calc.py#L282-L295

For some reason, `np.allclose` returns `True` in this case, even though the arrays have differing shapes:

```
In [6]: import numpy as np

In [7]: np.allclose(single_file.zsurf, ds.zsurf)
Out[7]: True
```

So when assignment is attempted in line 290, it fails, because the arrays have different shapes.  The workaround I've used now is to replace `np.allclose` with xarray's version, which does check the dimensions and shape of the arrays, but works as an assertion statement (rather than returning `True` or `False`).  Then assignment is no longer attempted (and everything works).

This workaround is OK, but it does not address the underlying issue, which is that *`zsurf` should not acquire a time dimension in the first place.*  I'm not aware of a way to prevent this with `xr.open_mfdataset`; with `xr.concat` this can be done by using the `data_vars='minimal'` keyword argument:

```
In [8]: xr.concat([xr.open_dataset('000{}0101.precip_monthly.nc'.format(year)) f
   ...: or year in [4, 5, 6]], dim='time', data_vars='minimal').zsurf
/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/conventions.py:393: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  result = decode_cf_datetime(example_value, units, calendar)
/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/conventions.py:412: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  calendar=self.calendar)
Out[8]:
<xarray.DataArray 'zsurf' (lat: 64, lon: 128)>
array([[ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       ...,
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.]])
Coordinates:
  * lon      (lon) float64 0.0 2.812 5.625 8.438 11.25 14.06 16.88 19.69 ...
  * lat      (lat) float64 -87.86 -85.1 -82.31 -79.53 -76.74 -73.95 -71.16 ...
Attributes:
    units:          m
    long_name:      Surface height
    time_avg_info:  average_T1,average_T2,average_DT
    cell_methods:   time: mean
```

My impression is that the ideal solution would be if we could provide an argument like `data_vars='minimal'` to `xr.open_mfdataset`, but that's not an available option.  Does that make sense?  Am I missing a way of doing this that already exists?

spencerahill:
Thanks for digging into this.

>For some reason, np.allclose returns True in this case, even though the arrays have differing shapes

I stumbled upon this myself in #181 (although I'm not sure I made a written note of it).  This sure seems like a bug to me: the [allclose documentation](https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html) only gives examples of same-shaped arrays, and there's an [open issue](https://github.com/numpy/numpy/issues/5654) at Numpy about basically the same thing.  But it's >2 years old with no movement, so we might consider giving it a ping.  

Here's the MWE I came up with; including version to show it still exists on latest numpy release:

```python
In [2]: np.version.version
Out[2]: '1.13.1'

In [5]: np.allclose(np.ones([5, 1]), np.ones([1]))
Out[5]: True
```

And there's some further odd behavior:

```python
In [7]: np.allclose(np.ones([5, 1]), np.ones([5, 2]))
Out[7]: True

In [8]: np.allclose(np.ones([5, 1]), np.ones([2]))
Out[8]: True

In [9]: np.allclose(np.ones([5, 1]), np.ones([]))
Out[9]: True
```

I would expect all of these to raise a `ValueError`, just like the following does:

```python
In [12]: np.allclose(np.ones([5, 1]), np.ones([12, 34]))
...
ValueError: operands could not be broadcast together with shapes (5,1) (12,34)
```

>The workaround I've used now is to replace np.allclose with xarray's version, which does check the dimensions and shape of the arrays.

I think that's a good decision even if there weren't this bug; we should strive to use xarray versions whenever possible (which we mostly already do anyways).

>My impression is that the ideal solution would be if we could provide an argument like data_vars='minimal' to xr.open_mfdataset, but that's not an available option. Does that make sense? Am I missing a way of doing this that already exists?

Is there a way to use the `preprocess` kwarg to accomplish this?  Feels a bit hackish though even if it could work.

spencerkclark:
> Is there a way to use the preprocess kwarg to accomplish this? Feels a bit hackish though even if it could work.

Since `preprocess` acts on individual datasets before they are combined, this might be tricky.  In looking over the code for [`xr.open_mfdataset`](https://github.com/pydata/xarray/blob/4a15cfa6f3e680af6969f3e2c6a375e2cc8c3288/xarray/backends/api.py#L433), I might be inclined to write our own custom version (leaving most of the logic the same), using something involving `xr.concat` / `xr.merge` in place of `xr.auto_combine` .

spencerahill:
>I might be inclined to write our own custom version (leaving most of the logic the same)

Probably worth at least asking at the xarray repo or mailing list before rolling our own.  I suspect others have/will come across the same problem.

spencerkclark:
@spencerahill I did a little more investigating -- it turns out we can obtain the desired behavior if we set grid attributes as coordinates before concatenating in `xr.open_mfdataset`.  Compare the following minimal examples:

### Concatenation with two variables (original behavior):
```python
for i in range(3):
    da = xr.DataArray(np.ones((2, 3)),
                      dims=['x', 't'],
                      coords=[np.arange(2), np.arange(3 * i, 3 * (i + 1))],
                      name='a')
    coord = xr.DataArray(2. * np.ones(2), dims=['x'], coords=[np.arange(2)], name='area')
    ds = da.to_dataset()
    ds['area'] = coord
    ds.to_netcdf('example-case-{:d}.nc'.format(i))
```
```xr.open_mfdataset('example-case-*.nc')``` returns:
```
<xarray.Dataset>
Dimensions:  (t: 9, x: 2)
Coordinates:
  * x        (x) int64 0 1
  * t        (t) int64 0 1 2 3 4 5 6 7 8
Data variables:
    a        (x, t) float64 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 ...
    area     (t, x) float64 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 ...
```
### Concatenation with a non-index coordinate (desired / current behavior):
```python
for i in range(3):
    da = xr.DataArray(np.ones((2, 3)),
                      dims=['x', 't'],
                      coords=[np.arange(2), np.arange(3 * i, 3 * (i + 1))],
                      name='a')
    coord = xr.DataArray(2. * np.ones(2), dims=['x'], coords=[np.arange(2)], name='area')
    da['area'] = coord
    da.to_netcdf('example-case-{:d}.nc'.format(i))
```

```xr.open_mfdataset('example-case-*.nc')``` returns:
```
<xarray.Dataset>
Dimensions:  (t: 9, x: 2)
Coordinates:
  * x        (x) int64 0 1
    area     (x) float64 2.0 2.0
  * t        (t) int64 0 1 2 3 4 5 6 7 8
Data variables:
    a        (x, t) float64 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 ...
```

Setting the time attributes as coordinates before calling `_prep_time_data` causes problems, so I hold off on setting those attributes until after calling `data_loader._prep_time_data`; the rest I set in our preprocessing function we pass to `xr.open_mfdataset`.  This solution could still probably use some cleanup, but I think it's a better approach to solving the underlying issue than writing our own `open_mfdataset` function.

File: aospy/data_loader.py [hunk 1]
```diff
@@ -91,7 +92,11 @@ def set_grid_attrs_as_coords(ds):
     Dataset
         Dataset with grid attributes set as coordinates
     """
-    int_names = GRID_ATTRS.keys()
+    if not set_time_vars:
```
spencerahill: Slight preference to use the affirmative, i.e. `if set_time_vars:` as the main `if` statement, with the `else` corresponding to `not set_time_Vars`

spencerahill:
Thanks @spencerkclark, nice find of a pretty simple workaround.

Just confirming, the only difference between the two examples is:

```python
ds = da.to_dataset()
ds['area'] = coord
ds.to_netcdf('example-case-{:d}.nc'.format(i))
```

versus

```python
da['area'] = coord
da.to_netcdf('example-case-{:d}.nc'.format(i))
```

I don't fully understand why this results in the behavior that it does...is this intuitive to you?  (I.e. should xarray consider tweaking or clarifying?)

>This solution could still probably use some cleanup, but I think it's a better approach to solving the underlying issue than writing our own open_mfdataset function.

It actually looks pretty tight already to me, although as I mentioned I'm not seeing it 100% clearly.  Need some tests of course; those might actually help me understand also.

spencerkclark:
@spencerahill yes, that is exactly the difference; it seems that `xr.auto_combine` treats data variables and coordinates differently.

> I don't fully understand why this results in the behavior that it does...is this intuitive to you? (I.e. should xarray consider tweaking or clarifying?)

No, this isn't super intuitive to me either; however, it is consistent with the [default behavior of `xr.concat`, which is to have `data_vars='all'` and `coords='different'`](https://github.com/pydata/xarray/blob/d1ab8051bd4ae42a2e197a7175db1aa915363107/xarray/core/combine.py#L15-L17) (and `xr.auto_combine` calls `xr.concat` underneath, with only default values for these arguments).  Perhaps this could be addressed by passing those arguments down through one's call to `xr.open_mfdataset` (so that one could have control over them)?

spencerahill:
Also FYI it turns out that the `np.allclose` behavior [isn't a bug](https://github.com/numpy/numpy/issues/5654)

spencerahill:
>Perhaps this could be addressed by passing those arguments down through one's call to xr.open_mfdataset (so that one could have control over them)?

Yes, that sounds like a good approach.

spencerkclark:
> Also FYI it turns out that the np.allclose behavior isn't a bug

That thread makes sense; thanks for double checking with the numpy folks on that.

> Yes, that sounds like a good approach.

Remarkably it seems someone else is keen on implementing the same thing: https://github.com/pydata/xarray/pull/1580.

spencerahill:
>Remarkably it seems someone else is keen on implementing the same thing

Haha go figure!

If that makes it into xarray 0.10, we have the option of waiting for that and then bumping our required xarray version to 0.10.  Conversely, we can continue with this PR using your workaround.  What do you think?

spencerkclark:
> If that makes it into xarray 0.10, we have the option of waiting for that and then bumping our required xarray version to 0.10. Conversely, we can continue with this PR using your workaround. What do you think?

I'll go ahead and continue with this PR.  If the change does make into xarray 0.10 then it's a simple change.  (This way we'll be prepared even if it doesn't).

spencerahill:
Great, in that case just ping me when you're ready for another review

spencerkclark:
@spencerahill great, this should be ready for another review when you get a chance.

spencerahill:
Looks good!  Thanks @spencerkclark

spencerahill:
https://github.com/pydata/xarray/pull/1580 has now been merged, although xarray v0.10 in which it will appear has not yet been released.

So if/when we come across something like this again, we should use the new xarray functionality and go ahead and bump the minimum requirement of xarray to >=0.10.

PR #200, opened 2017-09-01
Overwrite time array as average of time bounds if time bounds present

### LINKED ISSUE #185: Incorrect temporal grouping when data's time array matches the end-dates of its time_bounds array
[utils.times._assert_has_data_for_time](https://github.com/spencerahill/aospy/blob/develop/aospy/utils/times.py#L387) is meant to ensure that the data loaded from disk for a Calc actually contains data spanning the full range of times specified for that Calc.  It uses the `raw_start_date` and `raw_end_date` attributes of the data to perform this check if they're there.  These attributes are set by [ensure_time_avg_has_cf_metadata](https://github.com/spencerahill/aospy/blob/develop/aospy/utils/times.py#L306).  If they don't exist, it just pulls the first and last times from the array.

However, I have come across a case where this logic failed.  The case is monthly data whose netCDF files' time metadata is incorrect, creating an off-by-one error (i.e. the file corresponding to January of year 1 was being decoded to February of year 1, and so on).  Although this is easily fixed via the `input_time_offsets` kwarg to `submit_mult_calcs`, I didn't realize this at first, yet aospy didn't raise or warn.  Instead, the `raw_data_start_date` was set to what is *supposed* to be the start date, i.e. January of year 1, rather than the actual decoded start date, which is February of year 1.  Compare `raw_data_start_date` and `time` in the snippet of this repr:

```python
Coordinates:
  ...
  * time                 (time) datetime64[ns] 1678-02-01 1678-03-01 ...
    time_weights         (time) timedelta64[ns] 31 days 28 days 31 days ...
    raw_data_start_date  datetime64[ns] 1678-01-01
    raw_data_end_date    datetime64[ns] 1682-01-01
    subset_start_date    datetime64[ns] 1678-01-01
    subset_end_date      datetime64[ns] 1680-12-30
  ...
```

Since `_assert_data_for_time` uses `raw_data_{start,end}_date` if they're there, it erroneously doesn't raise.  @spencerkclark any thoughts on how best to fix this?

spencerahill:
Closes #185.

Maybe `_average_time_bounds` should go in `utils.times`?

Also still need:
~~- [ ] More informative message for the NotImplementedError~~
- [x] tests
- [x] what's new


aospy/data_loader.py
```diff
@@ -127,16 +128,54 @@ def _sel_var(ds, var):
...
+    ------
+    NotImplementedError
+        If the time bounds array doesn't match one of the supported shapes
+        specified above.
+
+    """
+    bounds = ds[TIME_BOUNDS_STR]
+    if bounds.ndim == 1:
+        assert len(bounds) == len(ds[TIME_STR]) + 1
+        avg = 0.5*(bounds[:-1] + bounds[1:])
```
spencerkclark: Would this work?  I think since `bounds` is indexed by a `'time_bounds'` coordinate, xarray will try to align `bounds[:-1]` with `bounds[1:]` so in this case `avg` would end up being the equivalent of `bounds.isel(time_bounds=slice(1, -1))` (which is not the desired result).

I think something like the following is more what we're after?
```python
avg = (0.5 * (bounds.shift(**{TIME_BOUNDS_STR: -1}) + bounds)).dropna(TIME_BOUNDS_STR)
```

E.g.
```python
In [1]: import numpy as np

In [2]: import xarray as xr

In [3]: example = xr.DataArray(np.arange(5), coords=[np.arange(5)], dims=['time_bounds'])

In [4]: example
Out[4]:
<xarray.DataArray (time_bounds: 5)>
array([0, 1, 2, 3, 4])
Coordinates:
  * time_bounds  (time_bounds) int64 0 1 2 3 4

In [5]: 0.5 * (example[1:] + example[:-1])
Out[5]:
<xarray.DataArray (time_bounds: 3)>
array([ 1.,  2.,  3.])
Coordinates:
  * time_bounds  (time_bounds) int64 1 2 3

In [6]: example.isel(time_bounds=slice(1, -1))
Out[6]:
<xarray.DataArray (time_bounds: 3)>
array([1, 2, 3])
Coordinates:
  * time_bounds  (time_bounds) int64 1 2 3

In [7]: (0.5 * (example.shift(time_bounds=-1) + example)).dropna('time_bounds')
Out[7]:
<xarray.DataArray (time_bounds: 4)>
array([ 0.5,  1.5,  2.5,  3.5])
Coordinates:
  * time_bounds  (time_bounds) int64 0 1 2 3
```

aospy/data_loader.py
```diff
@@ -127,16 +128,54 @@ def _sel_var(ds, var):
...
+    elif bounds.ndim == 2:
+        assert set([2, len(ds[TIME_STR])]) == set(bounds.shape)
```
spencerkclark: I think 
```python
assert bounds.sizes == {TIME_STR: ds.sizes[TIME_STR], BOUNDS_STR: 2}
```
might be slightly more idiomatic xarray (because it takes into account the dimension label-size correspondence).

aospy/data_loader.py
```diff
@@ -127,16 +128,54 @@ def _sel_var(ds, var):
...
+    Raises
+    ------
+    NotImplementedError
+        If the time bounds array doesn't match one of the supported shapes
+        specified above.
+
+    """
+    bounds = ds[TIME_BOUNDS_STR]
+    if bounds.ndim == 1:
+        assert len(bounds) == len(ds[TIME_STR]) + 1
```
spencerkclark: If this (or the other `assert` statement below) raises an `AssertionError`, should we provide a more informative error description?

spencerkclark (commented):
Thanks for putting this together.  I agree; I do think `_average_time_bounds` should go in `utils.times`.

If I understand things correctly, I *think* there might be an issue in the logic for the case when time bounds is a 1D array (see my suggestion below).  Other than that, there are just a few minor things.

aospy/data_loader.py
spencerkclark: Actually this should be:
```python
assert dict(bounds.sizes) == {TIME_STR: ds.sizes[TIME_STR], BOUNDS_STR: 2}
```
because `bounds.sizes` is an `OrderedDict` (so we should cast it as an ordinary `dict` first).

aospy/data_loader.py
spencerahill: You're right; for some reason I was thinking about these as if they were just numpy arrays.  Good catch!

aospy/data_loader.py
spencerahill: @spencerkclark thanks for the review.  All of your suggestions are good; will implement.

spencerahill:
I now realize that the 1-D time_bounds use case actually has never arisen for me.  I was conflating it with the same behavior in AM2.1 but for *lat and lon bounds*:

```
/archive/s1h/am2/am2clim_reyoi/gfdl.ncrc3-intel-prod/pp/atmos/ts/monthly/30yr/$ ncdump -h atmos.198301-201212.t_surf.nc
netcdf atmos.198301-201212.t_surf {
dimensions:
	time = UNLIMITED ; // (360 currently)
	lat = 90 ;
	latb = 91 ;
	lon = 144 ;
	lonb = 145 ;
	nv = 2 ;
variables:
	double average_DT(time) ;
		average_DT:long_name = "Length of average period" ;
		average_DT:units = "days" ;
		average_DT:missing_value = 1.e+20 ;
		average_DT:_FillValue = 1.e+20 ;
	double average_T1(time) ;
		average_T1:long_name = "Start time for average period" ;
		average_T1:units = "days since 1982-01-01 00:00:00" ;
		average_T1:missing_value = 1.e+20 ;
		average_T1:_FillValue = 1.e+20 ;
	double average_T2(time) ;
		average_T2:long_name = "End time for average period" ;
		average_T2:units = "days since 1982-01-01 00:00:00" ;
		average_T2:missing_value = 1.e+20 ;
		average_T2:_FillValue = 1.e+20 ;
	double lat(lat) ;
		lat:long_name = "latitude" ;
		lat:units = "degrees_N" ;
		lat:cartesian_axis = "Y" ;
		lat:edges = "latb" ;
	double latb(latb) ;
		latb:long_name = "latitude edges" ;
		latb:units = "degrees_N" ;
		latb:cartesian_axis = "Y" ;
	double lon(lon) ;
		lon:long_name = "longitude" ;
		lon:units = "degrees_E" ;
		lon:cartesian_axis = "X" ;
		lon:edges = "lonb" ;
	double lonb(lonb) ;
		lonb:long_name = "longitude edges" ;
		lonb:units = "degrees_E" ;
		lonb:cartesian_axis = "X" ;
	double nv(nv) ;
		nv:long_name = "vertex number" ;
		nv:units = "none" ;
		nv:cartesian_axis = "N" ;
	float t_surf(time, lat, lon) ;
		t_surf:long_name = "surface temperature" ;
		t_surf:units = "deg_k" ;
		t_surf:valid_range = 100.f, 400.f ;
		t_surf:missing_value = 1.e+20f ;
		t_surf:_FillValue = 1.e+20f ;
		t_surf:cell_methods = "time: mean" ;
		t_surf:time_avg_info = "average_T1,average_T2,average_DT" ;
	double time(time) ;
		time:long_name = "time" ;
		time:units = "days since 1982-01-01 00:00:00" ;
		time:cartesian_axis = "T" ;
		time:calendar_type = "NOLEAP" ;
		time:calendar = "NOLEAP" ;
		time:bounds = "time_bounds" ;
	double time_bounds(time, nv) ;
		time_bounds:long_name = "time axis boundaries" ;
		time_bounds:units = "days" ;
		time_bounds:missing_value = 1.e+20 ;
		time_bounds:_FillValue = 1.e+20 ;
```

So, unless somebody else has such a use case, I'm inclined to remove this logic, and only accept bounds arrays with shape `(len(times), 2)`.  That makes the function simpler and the testing much easier.  @spencerkclark what do you think?

spencerkclark:
> So, unless somebody else has such a use case, I'm inclined to remove this logic, and only accept bounds arrays with shape (len(times), 2). That makes the function simpler and the testing much easier.

I agree; I also have not encountered 1D time bounds, so this sounds like a good idea.

spencerahill:
Great.  Also FWIW, [CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-boundaries) only permit the 2D format:

>To represent cells we add the attribute bounds to the appropriate coordinate variable(s). The value of bounds is the name of the variable that contains the vertices of the cell boundaries. We refer to this type of variable as a "boundary variable." *A boundary variable will have one more dimension than its associated coordinate or auxiliary coordinate variable.* The additional dimension should be the most rapidly varying one, and its size is the maximum number of cell vertices.

spencerahill:
>CF conventions only permit the 2D format

Given this, I'm wondering if we even need this check at all.  It's hard to imagine an input dataset wherein the time bounds and time arrays are different lengths.  So I'm inclined to just remove the check entirely.

(Motivated by coveralls tipping me off that my test doesn't cover this mis-shaped case.)

aospy/utils/times.py
```diff
@@ -51,6 +55,46 @@ def apply_time_offset(time, years=0, months=0, days=0, hours=0):
...
+    bounds = ds[TIME_BOUNDS_STR]
+    if not dict(bounds.sizes) == {TIME_STR: ds.sizes[TIME_STR], BOUNDS_STR: 2}:
+        msg = ("Time bounds array has unsupported shape of {0}.  Time bounds "
+               "must have one dimension corresponding to the Dataset's time "
+               "array and one 'bounds' dimension of length 2.  Length of "
+               "Dataset's time array is {1}.").format(bounds.shape,
+                                                      len(ds[TIME_STR]))
+        raise ValueError(msg)
+
+    start_bounds = bounds.isel(**{BOUNDS_STR: 0})
```
spencerkclark: The function you've written here is nice, because it can be applied to both dates and floats.  That said, we only use this function on floats (i.e. before we decode times).  Therefore optionally I think it would be possible to simplify the logic (and tests) slightly:
```python
def average_time_bounds(ds):
    bounds = ds[TIME_BOUNDS_STR]
    new_times = bounds.mean(dim=BOUNDS_STR, keep_attrs=True).drop('time').rename(TIME_STR)
    new_times[TIME_STR] = new_times
    return new_times
```

aospy/data_loader.py
```diff
@@ -151,8 +152,9 @@ def _prep_time_data(ds):
     """
     ds = times.ensure_time_as_dim(ds)
     ds, min_year, max_year = times.numpy_datetime_workaround_encode_cf(ds)
-    if internal_names.TIME_BOUNDS_STR in ds:
+    if TIME_BOUNDS_STR in ds:
         ds = times.ensure_time_avg_has_cf_metadata(ds)
+        ds[TIME_STR].values = times.average_time_bounds(ds).values
```
spencerkclark: If you went with my suggestion for the simpler method below, I *think* you could simplify this to:
```python
ds[TIME_STR] = times.average_time_bounds(ds)
```

aospy/test/test_utils_times.py
```diff
@@ -530,5 +534,29 @@ def test_yearly_average_masked_data():
     xr.testing.assert_allclose(actual, desired)
 
 
-if __name__ == '__main__':
-    sys.exit(unittest.main())
+def test_average_time_bounds():
```
spencerkclark: Regardless of whether or not you switch to a float-only method, could you add a test for the use case where floats are used rather than dates, since that is the use case in `data_loader.py`?

spencerkclark (commented):
@spencerahill thanks for taking the time to switch `test_utils_times.py` to use `pytest`!  I have one optional suggestion that could simplify things a bit more, but it's totally up to you.  

Regarding checking the shape of `time_bounds`; I agree it is probably not necessary.

aospy/utils/times.py
spencerkclark: Note the slight edit above (the addition of `.drop`)

spencerahill:
>we only use this function on floats (i.e. before we decode times)

Wow, you're absolutely right.  Total mistake on my part.  Lesson: it pays to stop and think before you start hacking away.

Thanks for catching that.  I implemented your suggested replacement. and it seems to work; will push when tests are good.

spencerahill:
I took the liberty of creating a pytest.fixture for the new test and rewriting one existing test, `test_ensure_time_avg_has_cf_metadata`, to use it.  There's room for further improvement using fixtures in this module, if/when somebody has time/inclination.

spencerahill:
We're green.  @spencerkclark I'll wait for your final approval, given how many careless mistakes I've made so far in this PR! 😝

spencerkclark:
Thanks, this looks good to me!

Issue #208, opened 2017-09-19
Adding custom reduction methods


chuaxr:
For my current purposes, I am frequently interested in computing averages of a variable over some condition across the time and domain, treating each as an independent sample. For example, computing the average of vertical velocity over regions with precipitation exceeding a certain threshold. Since masking is required to isolate the samples of interest, stacking the data in (time, lat, lon) and then taking the mean (the 'stacked average') is not the same as taking the mean in time, then in (lat, lon).

One workaround is to simply define the stacked average as a user function. However, this leads to duplication of variables such as w_conditioned_on_precip and stacked_average_of_w_conditioned_on_precip. To avoid that, @spencerkclark helped me add a custom reduction method in my copy of aospy. 

It would be ideal for a user to add custom reduction methods as desired. If that is unlikely to happen in the near future, though, it would be nice if this 'stacked average' were added in the next release.

spencerahill:
I agree, supporting custom reduction methods would be extremely useful; @spencerkclark and I were recently talking offline about exactly that.

We could also consider adding a stacked average directly.  I don't immediately understand the operation as you've described it though...maybe you'd be willing to share e.g. your `w_conditioned_on_precip` function?

Initial thoughts re: supporting custom methods:

- Reductionsare currently handled in, [Calc._time_reduce](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L483-L497) for gridpoint-by-gridpoint and [Calc.region_calcs](https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L499-L529) for regional reductions. 

- In terms of how the user would specify a custom reduction, my first inclination is to just let them pass the function and its label into `calc_suite_specs['output_time_regional_reductions']`.  E.g. if you have a function `my_custom_reduction`, support 

```python
calc_suite_specs=dict(
    ..., 
    output_time_regional_reductions=['av', 'std', ('my_reduc', my_custom_reduction)],
    ...)
```

As the code currently stands, the function would have to accept an xr.DataArray as it's sole argument (and one in which the calculation and within-year averaging already performed).

- Another option would be to create a `Reduction` class.  E.g. starting out like

```python
class Reduction(object):
    def __init__(self, name, function):
        self.name = name
        self.function = function
```

[...]

PR #218, opened 2017-10-16
Cast float32 arrays to float64 arrays in _sel_var

### LINKED ISSUE #217: Inaccuracy of some operations (e.g. stacked average) when underlying data is float32
If no data is missing in the underlying dataset, then the stacked average (#208) and reg.av should yield the same answer. (In my case, doing a reg.av automatically results in a time average, since all data is output within the same year).

However, this is not the case.
 
Using the full data file (1000 time steps at 20s, ~5h on a 96x96 horizontal grid), I obtain the following:
regav_sav

With only the first hour of data, the difference isn't as large:
regav_sav_1h

Adding a typecast statement in stack_mean (see code below) appeared to fix that (see plot below), presumably because the model output is in float32 and reg.av does a typecast somewhere while the stacked average did not. Since any errors here contaminate all my analyses, I would very much appreciate knowing if there are other fixes I should make. For example, I may need to average over days at a very high frequency (e.g. 20s). In that case, would numpy float precision be enough, or would I have to average it within the model to make the size manageable?

```python
def stack_mean(self,var):
    """Takes mean of variable over stacked dimensions.
    Meant for masked data. 
                                                                                                             
    Parameters
    ---------
    var: DataArray to take mean of
    var_like: DataArray with appropriate time dimension
                                                                                                             
    """
    var=var.astype(float)
    var=var.stack(sdim=[internal_names.TIME_STR,internal_names.LAT_STR,internal_names.LON_STR]).mean('sdim')
    return(var)

```
Same as the first plot, except with modified stack_mean
regav_sav_5h_float

spencerkclark:
Closes #217 

@spencerahill (cc: @chuaxr) here's a proposed solution for #217.  If you're on board with it, let me know where might be best for the warning in the documentation.


spencerahill:
Thanks @spencerkclark.  LGTM so far -- still needs a kwarg-based flag that the user can use to turn this off if they want.

For the documentation, [here](https://aospy.readthedocs.io/en/latest/examples.html#runs-and-dataloaders) seems like a good place for a Note on this.  It should also appear in the API reference page, maybe as a Note at the top of the [DataLoader section](https://aospy.readthedocs.io/en/latest/api.html#aospy.data_loader.DataLoader.load_variable).  And the on/off flag kwarg will need docs + tests + what's new mention.

chuaxr:
Thanks @spencerkclark, I've tried out the updated data_loader.py and it does address my issue.

spencerkclark:
@spencerahill whenever you get a chance, this should be ready for another review.

spencerahill:
@spencerkclark great, tied up most of today but should manage by the end of the week.

aospy/data_loader.py
```diff
@@ -103,7 +103,30 @@ def set_grid_attrs_as_coords(ds, set_time_vars=True):
...
+                        'For accurate reduction operations using bottleneck, '
+                        'datapoints are being cast to the np.float64 datatype.'
+                        ' For more information see: https://github.com/pydata/'
+                        'xarray/issues/1346')
+        return da.astype(np.float64)
+    else:
+        return da
+
+
+def _sel_var(ds, var, upcast_float32):
```
spencerahill: Here and elsewhere, I think it would make more sense for `upcast_float32` to be a keyword argument rather than a positional one.  That makes it's role as a flag to change the logic more clear.  With default True.

spencerahill: Sorry, I see now that elsewhere it is a kwarg.  So just make it a kwarg here too.

aospy/data_loader.py
```diff
@@ -411,6 +452,7 @@ class GFDLDataLoader(DataLoader):
     """
     def __init__(self, template=None, data_direc=None, data_dur=None,
                  data_start_date=None, data_end_date=None,
+                 upcast_float32=None,
```
spencerahill, line 455: Why `None` here but `True` in the other data loaders?  I think `True` would work here also.

docs/api.rst
```diff
@@ -64,6 +64,17 @@ Currently supported data loader types are :py:class:`DictDataLoader`,
...
+
+   An important consideration can be the datatype used to store values in
+   one's datasets.  In particular, if the float32 datatype is used in
+   storage, it can lead to undesired inaccuracies in the computation of
+   reduction operations (like means) due to upstream issues.  To address
+   this it is recommended to always upcast float32 data to float64.  This
+   behavior is turned on by default.  If one would like to disable this
+   behavior one can set the ``upcast_float32`` argument in their ``DataLoader``
+   constructors to ``False``.
+
```
spencerahill: Let's add a link to the xarray issue in these notes, as a parenthetical after "upstream issues", as you did in the What's New.

Also, let's switch these to be the second person imperative, e.g. 'your dataset', and 'You can disable this behavior by setting the ...'.  That makes it feel more inviting.

spencerahill (commented):
A few minor comments, mostly stylistic.  Nice work @spencerkclark per usual.  Feel free to merge once they're addressed and CI passes; no need for me to review again.  Thanks!

aospy/data_loader.py
```diff
@@ -439,12 +481,20 @@ def __init__(self, template=None, data_direc=None, data_dur=None,
                 self.preprocess_func = preprocess_func
             else:
                 self.preprocess_func = template.preprocess_func
+            if upcast_float32 is not None:
```
spencerkclark, line 484: > Why None here but True in the other data loaders? I think True would work here also.

The distinction is that to support the construction of a `GFDLDataLoader` via a template `GFDLDataLoader` we need a neutral default value.  This allows us to unambiguously identify if the user wants to use value from the template, or override it. 

Consider the case where `upcast_float32` was `False` in the template.  If we used a default value of `True` we would not know if the user simply did not specify an `upcast_float32` argument (indicating that they would want to use the template's value) or wanted to override it (i.e. instead of `False` use `True`).

aospy/data_loader.py
```diff
@@ -439,12 +481,20 @@ def __init__(self, template=None, data_direc=None, data_dur=None,
...
+                self.upcast_float32 = upcast_float32
+            else:
+                self.upcast_float32 = template.upcast_float32
         else:
             self.data_direc = data_direc
             self.data_dur = data_dur
             self.data_start_date = data_start_date
             self.data_end_date = data_end_date
             self.preprocess_func = preprocess_func
+            if upcast_float32 is None:
```
spencerkclark, line 494: In the case that a template `GFDLDataLoader` is not provided, we translate the default argument of `None` to mean `True`.

spencerkclark (commented):
@spencerahill thanks for the review!  I will push some changes soon.  I have a couple minor clarification comments as to why I used `None` as the default value for `upcast_float32`; let me know if those make sense.

spencerkclark:
@spencerahill we're green here, but I just want to double check that you're cool with my above comments before merging.

spencerahill:
@spencerkclark thanks for the clarification.  OK, that's fine by me.  I'll go ahead and merge.  The Travis failure is another instance of #219.

Thanks for taking care of this so quickly!  So @chuaxr, if you install from our develop branch (rather than pip or conda), this should be working now that this is merged.

chuaxr:
@spencerahill I'm waiting for #208 to be resolved before doing a full update. In the meantime I've used data_loader.py from this commit, which should serve as a temporary workaround.

PR #220, opened 2017-10-20
Limit number of workers in external_client in test_automate.py

### LINKED ISSUE #219: TimeoutError in  TravisCI
For the past week or so, we are now always getting a failure in exactly one Travis CI environment, but not always the same one -- I've seen in the 2.7, 2.7-min, and 3.6 ones so far.  These all are TimoutErrors in the ` test_submit_two_calcs_external_client` test.  See e.g. https://travis-ci.org/spencerahill/aospy/jobs/288531602

This feels like just something buggy on the Travis CI end, but it's still annoying.  Should look into and fix if we can.  In the meantime this issue serves at least as a heads up.

spencerkclark:
Closes #219


spencerkclark:
@spencerahill hmm this just seems to be kicking the can down the road.  Now instead of a TimeoutError in distributed we get one on Travis: https://travis-ci.org/spencerahill/aospy/jobs/290504820

spencerkclark:
I feel like my last commit was the key one.  That reduced the total build time to 20 minutes and 45 seconds (across all five builds on Travis).  Prior successful builds were taking a total time on the order of 31 minutes.

I'll roll back my changes specifying a larger `death_timeout`, because I think they are likely not necessary.

spencerahill:
@spencerkclark thanks for fixing this so quickly!  Glad that it didn't end up being a quagmire.

PR #222, opened 2017-10-22
WIP got rid of Units class

micahkim23:
@spencerahill How does this look?


docs/api.rst
```diff
@@ -122,6 +122,12 @@ Var
 
     .. automethod:: aospy.var.Var.__init__
 
+.. note::
+
+   There has been discussion of implementing units-handling upstream
```
spencerkclark: I realize you're just copying this text from the old documentation for `Units` objects (which is totally fine), but I think it might be good to take the opportunity to be a little more descriptive here.  Maybe something along the lines of:
```
.. note::

    While for the sake of tracking metadata we encourage users to add a ``units`` 
    attribute to each of their :py:class:`Var` objects, these ``units`` attributes provide 
    nothing more than descriptive value.  One day we hope DataArrays 
    produced by loading or computing variables will be truly units-aware 
    (e.g. adding or subtracting two DataArrays with different units will lead to 
    an error, or multiplying two DataArrays will result in a DataArray with new units), 
    but we will leave that to upstream libraries to implement
    (see `pydata/xarray#525 <https://github.com/pydata/xarray/issues/525>`_ 
    for more discussion).
```

spencerkclark (commented):
@micahkim23 this looks great!  Thanks for putting this together.  I just have one comment within the diff for the documentation.

It would also be great if you could add a note in `whats-new.rst` regarding this change.  As @spencerahill suggests (https://github.com/spencerahill/aospy/issues/50#issuecomment-338227887), please update the version number for the "unreleased" heading to `0.3`.  I think this would probably be best noted under a new "Breaking changes" heading (since this will require some users to modify their object libraries).

docs/api.rst
micahkim23: @spencerkclark thanks for the suggestion, I'll fix that.

spencerahill: I agree, and just add 'the' between 'for' and 'sake'.

spencerahill:
Thanks @micahkim23.  Looks good overall.  I agree w/ @spencerkclark 's comments.  Once those are addressed this should be good to go!

micahkim23:
I was just wondering, since this change is supposed to be part of `0.3`, should the pull request not be made on the develop branch? Will it be a problem for code from `0.3` and `0.2.1` to be mixed?

spencerahill:
There's actually nothing that is formally binding us to v0.2.1 at this point.  Just change the `v0.2.1` to `v0.3` within the What's New, and then whenever I do the next release I will update everything (e.g. within `setup.py` and on our conda-forge feedstock) to v0.3 instead of v0.2.1.  Only once a release is actually performed does any code get formally tied to it.

docs/whats-new.rst
```diff
@@ -3,10 +3,17 @@
...
-v0.2.1 (unreleased)
--------------------
+v0.3 (unreleased)
+-----------------
+
+Breaking Changes
+~~~~~~~~~~~~~~~~
+
+- Deprecate ``Units`` class, so now the ``units`` attribute of the
+  ``Var`` class is a ``string``. (fixes :issue:`50` via pull:`222`).
```
spencerkclark: - Remove the double backticks from "string" (in other words make it ordinary text rather than a literal, because the literal name of the datatype is `str`)
- Add a colon before "pull"

spencerkclark (commented):
This looks good to me; just a couple minor nits!

spencerahill:
In it goes!  Thanks @micahkim23 for the heavy lifting and @spencerkclark for the review.

PR #223, opened 2017-10-30
Deprecate Constant class

### LINKED ISSUE #50: Remove obsolete Constant and Units classes

micahkim23:
Closes #50


aospy/calc.py
```diff
@@ -10,14 +10,18 @@
...
 
-from .constants import Constant, grav
 from . import internal_names
 from . import utils
 from .var import Var
 
 
 logging.basicConfig(level=logging.INFO)
 
+GRAV = (
```
spencerkclark: @micahkim23 @spencerahill would we like to make these DataArrays rather than tuples (as suggested in https://github.com/spencerahill/aospy/issues/50#issuecomment-299206776), or is it not worth it?  This would give the benefit of named attributes, and remove the need to select the first element when doing mathematical operations.

spencerkclark (commented):
Thanks for starting on this @micahkim23!  I have a couple initial comments on these changes.

It would be good to get @spencerahill's opinion on this too, but I think it might be a good idea to keep a module called `constants.py` to house just the minimum remaining necessary constants (e.g. `GRAV` and `R_E`).  One could import them from `calc.py`, `model.py`, and  `vertcoord.py`, and it would prevent the need to define `GRAV` twice.

aospy/calc.py
micahkim23: @spencerkclark I think the DataArrays are better than tuples. Later on, having tuples means that we will have to keep track of which indices we're selecting from the tuples which is cumbersome.

micahkim23:
I think it would be a good idea to have a `constants.py` file. In the future, we may need more constants and we don't want those to be defined all over the place.

aospy/calc.py
spencerahill: +1 on DataArrays.  I had been thinking just floats but DataArrays much better.

spencerahill:
+1 on a constants module, but I'm leaning towards adding a leading underscore, i.e. `_constants.py`, to signify that it's not intended as public API.

Although, especially now that some of my work involves atmospheres of other planets, it occurs to me that we should eventually be more flexible with the values of these constants.  But I think we can come back to that later as its own issue.

aospy/model.py
```diff
@@ -4,10 +4,14 @@
 import numpy as np
 import xarray as xr
 
-from .constants import r_e
 from . import internal_names
 from . import utils
 
+R_E = (
```
spencerahill: Let's rename this `RADIUS_EARTH` to be more explicit

aospy/calc.py
spencerahill: Let's rename `GRAV_EARTH` to be more explicit, c.f. my comments on `R_E` and about earth v. other planetary atmospheres.

docs/whats-new.rst
```diff
@@ -12,8 +12,8 @@ Breaking Changes
 ~~~~~~~~~~~~~~~~
 
 - Deprecate ``Units`` class, so now the ``units`` attribute of the
-  ``Var`` class is a string. (fixes :issue:`50` via :pull:`222`).
-  By `Micah Kim <https://github.com/micahkim23>`_.
+  ``Var`` class is a string. Deprecate ``Constant`` class. (fixes :issue:`50`
```
spencerahill: Make this a separate bullet point from the previous one about `Units`.

spencerahill:
LGTM besides a few minor comments I made above.  Thanks @micahkim23 for the PR and @spencerkclark for the review.

spencerkclark:
+1 for the the leading underscore for the `_constants.py` module.

> Although, especially now that some of my work involves atmospheres of other planets, it occurs to me that we should eventually be more flexible with the values of these constants. But I think we can come back to that later as its own issue.

It feels most natural for these constants to be settable attributes of `Run` objects, but I agree this would require a bit of discussion to figure out how it should be done best.

aospy/calc.py
```diff
@@ -470,7 +468,7 @@ def _compute_full_ts(self, data, monthly_mean=False, zonal_asym=False):
                                                  self.end_date)
             )
             if self.dtype_out_vert == 'vert_av':
-                full_ts *= (grav.value / self._to_desired_dates(self._ps_data))
+                full_ts *= (GRAV_EARTH.data / self._to_desired_dates(self._ps_data))
```
spencerkclark: Here and elsewhere the constants are used, we should not need to use the `data` attribute.  In other words I think the following should be OK:
```python
full_ts *= (GRAV_EARTH / self._to_desired_dates(self._ps_data))
```

docs/api.rst
```diff
@@ -192,20 +192,10 @@ automate
 Constants
```
spencerkclark: I'll let @spencerahill weigh in, but I think since we are denoting `_constants.py` as private API now, we should probably delete this section from the documentation entirely.

docs/whats-new.rst
```diff
@@ -11,6 +11,9 @@ v0.3 (unreleased)
 Breaking Changes
 ~~~~~~~~~~~~~~~~
 
+- Deprecate ``Constant`` class, so now we have a ``_constants.py`` file
+  containing commonly used constants. (fixes :issue:`50` via :pull:`223`).
```
spencerkclark: nit: let's try and be a little more specific, something along the lines of "... so now we have a ``_constants.py`` module containing only the physical constants required by ``aospy``, stored as scalar DataArrays."

aospy/_constants.py
```diff
@@ -0,0 +1,14 @@
+"""commonly used physical constants"""
```
spencerkclark: nit: let's say "Physical constants used in aospy", because this is by no means an exhaustive list of commonly used physical constants :)

spencerkclark (commented):
@micahkim23 I added a few more minor comments, but this is looking good!

spencerkclark:
@micahkim23 quick procedural question for you -- are you always squashing your commits before you push new changes?  The reason I ask is that I am not notified via email when you make updates (otherwise I would have provided my comments sooner!).  GitHub has a really nice ["Squash and merge"](https://github.com/blog/2141-squash-your-commits) option for PR's, so we can easily handle cleaning up the commits/commit messages upon merging.  This way you can just push your newest sets of commits as you go (not worrying about cluttering up the commit history), and @spencerahill and I will be notified immediately.

micahkim23:
yeah, I've been squashing my commits before pushing new changes. I won't do that anymore, so that you guys can get notified sooner. Thanks for the heads up.

docs/api.rst
spencerahill: I agree.

docs/whats-new.rst
spencerahill: Re-word as "Deprecate ``Constant`` class and ``constants.py`` module.  Physical constants used internally by aospy are now stored in ``_constants.py``  (fixes :issue:`50` via :pull:`223`)."

spencerkclark (approved):
Nice work @micahkim23.  The build failure for Python 3.4 on Windows looks unrelated.

spencerahill:
In it goes!  Thanks @micahkim23 for the code and @spencerkclark for the review.

As I mentioned to @micahkim23 offline, this may not seem like all that important, if nobody is using this functionality much anyways.  But in fact cleanup like this is really useful, as it helps us have a more streamlined codebase and enables us to think more clearly about the code that remains.

Issue #236, opened 2017-11-18
distributed.scheduler.KilledWorker for a long time series


chuaxr:
I currently wish to perform averages of model variables (say condensates) on a 100-days worth of data (~100x100x50 at hourly resolution) in 10 day blocks for a base experiment, as well as 10 day blocks for different experiments. Each 10 day chunk is in its own file. 

When attempting to perform an average over 6 variables on the 10 files in the 100 day run using parallelize = True, I obtain the following error: 

```
Traceback (most recent call last):
  File "aospy_main.py", line 180, in <module>
    calcs = submit_mult_calcs(calc_suite_specs, calc_exec_options)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/aospy/automate.py", line 460, in submit_mult_calcs
    return _exec_calcs(calcs, **exec_options)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/aospy/automate.py", line 306, in _exec_calcs
    return _submit_calcs_on_client(calcs, client, func)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/aospy/automate.py", line 265, in _submit_calcs_on_client
    return db.from_sequence(calcs).map(func).compute()
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/dask/base.py", line 99, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/dask/base.py", line 206, in compute
    results = get(dsk, keys, **kwargs)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/distributed/client.py", line 1949, in get
    results = self.gather(packed, asynchronous=asynchronous)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/distributed/client.py", line 1389, in gather
    asynchronous=asynchronous)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/distributed/client.py", line 559, in sync
    return sync(self.loop, func, *args, **kwargs)
  File "/nbhome/xrc/anaconda2/envs/py361/lib/python3.6/site-packages/distributed/utils.py", line 244, in sync
    six.reraise(*error[0])
[...]

PR #240, opened 2017-11-23
Expose data_vars and coords arguments to xr.open_mfdataset in DataLoaders

spencerkclark:
This option ultimately did make it in to xarray version 0.10.0, so we can now clean up the logic required to work around the issue discussed here https://github.com/spencerahill/aospy/pull/199#issuecomment-326752960.


spencerahill:
Thanks @spencerkclark!  Looks good to me.  Go ahead and merge at your discretion.

spencerkclark:
@spencerahill great!  I'm considering making `data_vars` and `coords` user-settable options in DataLoaders in light of #236 (probably within this PR).  Do you think that would be a good idea?

spencerahill:
@spencerkclark yes definitely.  Thanks for the in-depth troubleshooting on that one.

Ping me once that's in and ready and I'll give it another look over

spencerkclark:
@spencerahill @micahkim23 this should be ready for another review.

aospy/data_loader.py
```diff
@@ -89,11 +88,7 @@ def set_grid_attrs_as_coords(ds, set_time_vars=True):
...
     """
-    if set_time_vars:
-        int_names = GRID_ATTRS.keys()
-    else:
-        int_names = (set(GRID_ATTRS.keys()) -
-                     set(TIME_VAR_STRS))
+    int_names = GRID_ATTRS.keys()
     grid_attrs_in_ds = set(int_names).intersection(set(ds.coords) |
                                                    set(ds.data_vars))
     ds.set_coords(grid_attrs_in_ds, inplace=True)
```
spencerahill: Would you mind taking care of #244 while you're at it by removing this use of `inplace`?

aospy/data_loader.py
```diff
@@ -89,11 +88,7 @@ def set_grid_attrs_as_coords(ds, set_time_vars=True):
     Dataset
         Dataset with grid attributes set as coordinates
     """
-    if set_time_vars:
-        int_names = GRID_ATTRS.keys()
-    else:
-        int_names = (set(GRID_ATTRS.keys()) -
-                     set(TIME_VAR_STRS))
+    int_names = GRID_ATTRS.keys()
```
spencerahill: Could just use `GRID_ATTRS.keys()` directly rather than defining `int_names`

aospy/data_loader.py
```diff
@@ -450,49 +473,44 @@ class GFDLDataLoader(DataLoader):
...
-                self.upcast_float32 = True
-            else:
-                self.upcast_float32 = upcast_float32
+            self._set_default('upcast_float32', upcast_float32, True)
+            self._set_default('data_vars', data_vars, 'minimal')
+            self._set_default('coords', coords, 'minimal')
+            self._set_default('preprocess_func', preprocess_func,
+                              lambda ds, **kwargs: ds)
+
+    def _maybe_inherit(self, template, attr, value):
```
spencerahill: Since this and `_set_default` aren't using any of the object's methods or attributes, I think it would be cleaner to define these at the module level as functions: `def _maybe_inherit(obj, template, attr, value):` etc.

Also, are the two different enough to warrant separate definitions?  Am I correct that `_set_default(obj, attr, value, getattr(template, attr)) == _maybe_inherit(obj, template, attr, value)`?

If you do keep both, I'd suggest making the signatures more homogenous: obj, attr, value, template/default.

spencerahill (commented):
A couple minor points + one request.  Otherwise looks good!

aospy/data_loader.py
spencerkclark: Indeed!  Good catch.  I'll clean these up when I get a chance.

aospy/data_loader.py
spencerkclark: Sure thing, that should be straightforward.

aospy/data_loader.py
spencerkclark: I think this was the way it was prior to #199 (I think because of line length in the following line).  Happy to change it though.

spencerkclark:
Appveyor seems to have been well-behaved today :).  @spencerahill let me know if my changes look good to you.

spencerahill:
>Appveyor seems to have been well-behaved today

Quick, let's merge before it gets tempermental! 😜 

Thanks @spencerkclark , this is great.

PR #242, opened 2017-11-27
WIP skip over ValueError in _time_reduce

### LINKED ISSUE #202: Skip dtype_out_time options that cause errors?
Short term issue
-----------------

@chuaxr has a use case for computed variables that do not have a time dimension (i.e. their `def_time` attribute is `False`).  She also has a use case for variables that do.  It would be nice if she could submit calculations in the main script for both of these variables at the same time, with specifications to compute no time reduction (e.g. `dtype_out_time=None`) and specifications to compute a time reduction (e.g. `dtype_out_time='av'`), without having things halted by an error.  More explicitly say one has two variables named `without_time_dim` and `with_time_dim`, and wants to compute them for a single `Run` and specifies `dtype_out_time=['av', None]`; the main script would create two Calc objects:

The first would seek to compute:
- **`without_time_dim` with a time average**
- `without_time_dim` with no time reduction

The second would seek to compute:
- `with_time_dim` with a time average
- `with_time_dim` with no time reduction

The one in bold currently raises an error, and crashes the set of computations.  We ignore errors on Calc objects themselves, so the calculations for the `with_time_dim` variable would be fine, but since `without_time_dim` with a time average would crash, aospy would not get to computing `without_time_dim` with no time reduction.  For consistency and convenience, it seems like it could be a good idea report, yet ignore the exception caused by attempting to compute `without_time_dim` with a time average (rather than error) in a similar manner to what is done for Calc objects:

https://github.com/spencerahill/aospy/blob/ce7c78418811c9e0ffed7bc647f8df53c34e03ee/aospy/automate.py#L244-L256

Longer term thoughts
----------------------
This is a very specific issue caused by the current main-script / Calc pipeline approach, which might have a relatively simple fix, but it could also serve as motivation to explore ways to use aospy outside of this strict paradigm (e.g. it does not always make sense to compute the Cartesian product of all variables and options; right now I address this by creating separate main scripts, but I feel like there could be a better way to do this).

micahkim23:
Closes #202

### DIFF
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -10,7 +10,7 @@ import logging
 import pprint
 import traceback
 
-from .calc import Calc, CalcInterface
+from .calc import Calc, CalcInterface, _TIME_DEFINED_REDUCTIONS
 from .region import Region
 from .var import Var
 
@@ -22,7 +22,6 @@ _RUNS_STR = 'runs'
 _REGIONS_STR = 'regions'
 _VARIABLES_STR = 'variables'
 _TAG_ATTR_MODIFIERS = dict(all='', default='default_')
-_TIME_DEFINED_REDUCTIONS = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts']
 
 
 def _get_attr_by_tag(obj, tag, attr_name):
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -22,6 +22,7 @@ _RUNS_STR = 'runs'
 _REGIONS_STR = 'regions'
 _VARIABLES_STR = 'variables'
 _TAG_ATTR_MODIFIERS = dict(all='', default='default_')
+_TIME_DEFINED_REDUCTIONS = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts']
 
 
 def _get_attr_by_tag(obj, tag, attr_name):
@@ -237,8 +238,30 @@ class CalcSuite(object):
 
     def create_calcs(self):
         """Generate a Calc object for each requested parameter combination."""
+        specs = self._combine_core_aux_specs()
+
+        for spec in specs:
+            spec['dtype_out_time'] = _prune_invalid_time_reductions(spec)
+
         return [Calc(CalcInterface(**sp)) for sp in
-                self._combine_core_aux_specs()]
+                specs]
+
+
+def _prune_invalid_time_reductions(spec):
+    valid_reductions = []
+    if not spec['var'].def_time:
+        for reduction in spec['dtype_out_time']:
+            if reduction not in _TIME_DEFINED_REDUCTIONS:
+                valid_reductions.append(reduction)
+            else:
+                msg = ("Var {0} has no time dimension "
+                       "for the given time reduction "
+                       "{1} so this calculation will "
+                       "be skipped".format(spec['var'].name, reduction))
+                logging.info(msg)
+    else:
+        valid_reductions = spec['dtype_out_time']
+    return valid_reductions
 
 
 def _compute_or_skip_on_error(calc, compute_kwargs):
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -248,6 +248,7 @@ class CalcSuite(object):
 
 
 def _prune_invalid_time_reductions(spec):
+    """Prune time reductions of spec with no time dimension."""
     valid_reductions = []
     if not spec['var'].def_time:
         for reduction in spec['dtype_out_time']:
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -249,7 +249,7 @@ class CalcSuite(object):
 def _prune_invalid_time_reductions(spec):
     """Prune time reductions of spec with no time dimension."""
     valid_reductions = []
-    if not spec['var'].def_time and spec['dtype_out_time'] != None:
+    if not spec['var'].def_time and spec['dtype_out_time'] is not None:
         for reduction in spec['dtype_out_time']:
             if reduction not in _TIME_DEFINED_REDUCTIONS:
                 valid_reductions.append(reduction)
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -249,7 +249,7 @@ class CalcSuite(object):
 def _prune_invalid_time_reductions(spec):
     """Prune time reductions of spec with no time dimension."""
     valid_reductions = []
-    if not spec['var'].def_time:
+    if not spec['var'].def_time and spec['dtype_out_time'] != None:
         for reduction in spec['dtype_out_time']:
             if reduction not in _TIME_DEFINED_REDUCTIONS:
                 valid_reductions.append(reduction)
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -243,8 +243,7 @@ class CalcSuite(object):
         for spec in specs:
             spec['dtype_out_time'] = _prune_invalid_time_reductions(spec)
 
-        return [Calc(CalcInterface(**sp)) for sp in
-                specs]
+        return [Calc(CalcInterface(**sp)) for sp in specs]
 
 
 def _prune_invalid_time_reductions(spec):
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -174,12 +174,12 @@ class CalcInterface(object):
             dtype_in_vert=self.dtype_in_vert,
             dtype_in_time=self.dtype_in_time, intvl_out=self.intvl_out)
 
-        if not var.def_time:
-            for reduction in dtype_out_time:
+        if not self.def_time:
+            for reduction in self.dtype_out_time:
                 if reduction in _TIME_DEFINED_REDUCTIONS:
                     msg = ("Var {0} has no time dimension "
                            "for the given time reduction "
-                           "{1}".format(var.name, reduction))
+                           "{1}".format(self.name, reduction))
                     raise ValueError(msg)
 
 
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -180,7 +180,7 @@ class CalcInterface(object):
                     msg = ("Var {0} has no time dimension "
                            "for the given time reduction "
                            "{1}".format(var.name, reduction))
-                    logging.info(msg)
+                    raise ValueError(msg)
 
 
 class Calc(object):
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -492,7 +492,6 @@ class Calc(object):
         if self.dtype_in_time == 'av':
             return arr
         reductions = {
-            'None': lambda xarr: xarr,
             'ts': lambda xarr: xarr,
             'av': lambda xarr: xarr.mean(internal_names.YEAR_STR),
             'std': lambda xarr: xarr.std(internal_names.YEAR_STR),
--- a/aospy/calc.py
+++ b/aospy/calc.py
@@ -39,6 +39,7 @@ ps = Var(
     def_lat=True,
     def_lon=True,
 )
+_TIME_DEFINED_REDUCTIONS = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts']
 
 
 class CalcInterface(object):
@@ -172,6 +173,14 @@ class CalcInterface(object):
             domain=self.domain, intvl_in=self.intvl_in,
             dtype_in_vert=self.dtype_in_vert,
             dtype_in_time=self.dtype_in_time, intvl_out=self.intvl_out)
+
+        if not var.def_time:
+            for reduction in dtype_out_time:
+                if reduction in _TIME_DEFINED_REDUCTIONS:
+                    msg = ("Var {0} has no time dimension "
+                           "for the given time reduction "
+                           "{1}".format(var.name, reduction))
+                    logging.info(msg)
 
 
 class Calc(object):
--- a/aospy/test/data/objects/examples.py
+++ b/aospy/test/data/objects/examples.py
@@ -47,11 +47,9 @@ example_proj = Proj(
     models=[example_model]
 )
 
-precip_largescale = Var(
-    name='precip_largescale',
-    alt_names=('condensation_rain',),
+var_not_time_defined = Var(
+    name='var_no_time_def',
     def_time=False,
-    description='Precipitation generated via grid-scale condensation',
 )
 
 condensation_rain = Var(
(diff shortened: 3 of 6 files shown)


micahkim23:
@spencerahill @spencerkclark Let me know what you guys think about the pr.

aospy/calc.py
```diff
@@ -173,6 +174,14 @@ def __init__(self, proj=None, model=None, run=None, ens_mem=None, var=None,
             dtype_in_vert=self.dtype_in_vert,
             dtype_in_time=self.dtype_in_time, intvl_out=self.intvl_out)
 
+        if not var.def_time:
+            for reduction in dtype_out_time:
+                if reduction in _TIME_DEFINED_REDUCTIONS:
+                    msg = ("Var {0} has no time dimension "
+                           "for the given time reduction "
+                           "{1}".format(var.name, reduction))
+                    logging.info(msg)
```
spencerkclark: Per https://github.com/spencerahill/aospy/issues/202#issuecomment-348496875, let's raise an exception here instead of just logging a message (probably a `ValueError` is appropriate here).

spencerkclark (commented):
@micahkim23 this looks great so far!  It would be great if you could add some tests for this new logic.  

I would probably start by creating a variable that was not defined in time in our test object library: https://github.com/spencerahill/aospy/blob/develop/aospy/test/data/objects/examples.py.  Then I would write tests to see if the time reductions were skipped when using `CalcSuite` and that new variable, and if an error was raised if a standalone `Calc` object was created using that variable that included time reductions in `dtype_out_time`.  The tests would probably be best placed in `test_automate.py` and `test_calc_basic.py`.

aospy/calc.py
micahkim23: for testing, are the only possible values for `dtype_out_time`: `None`, some subset of `['av', 'std, 'ts', 'reg.av', 'reg.std', 'reg.ts', 'None']`, or individual strings of: `'av', 'std, 'ts', 'reg.av', 'reg.std', 'reg.ts', 'None'`?

aospy/automate.py
```diff
@@ -237,8 +238,30 @@ def _combine_core_aux_specs(self):
 
     def create_calcs(self):
         """Generate a Calc object for each requested parameter combination."""
-        return [Calc(CalcInterface(**sp)) for sp in
-                self._combine_core_aux_specs()]
+        specs = self._combine_core_aux_specs()
+
+        for spec in specs:
+            spec['dtype_out_time'] = _prune_invalid_time_reductions(spec)
```
micahkim23, line 243: is `spec['dtype_out_time']` always a tuple? I'm confused because in here https://github.com/spencerahill/aospy/blob/develop/aospy/calc.py#L153-L156, it's possible for `dtype_out_time` to be a string, but then it's changed back to a tuple.

aospy/calc.py
spencerkclark: Yes 👍

aospy/automate.py
spencerkclark: I think it's reasonably safe to say here that it will be an iterable of some kind; the documentation states that a list is required in the `calc_suite_specs` dictionary:
http://aospy.readthedocs.io/en/latest/api.html?highlight=calcsuite#aospy.automate.submit_mult_calcs.

aospy/test/data/objects/examples.py
```diff
@@ -47,6 +47,13 @@ def total_precipitation(convection_rain, condensation_rain):
     models=[example_model]
 )
 
+precip_largescale = Var(
```
spencerahill: Why is this new `Var` object needed?

aospy/test/test_automate.py
```diff
@@ -351,3 +354,25 @@ def test_permute_aux_specs(self, calc_suite, calcsuite_specs):
         assert len(actual) == len(expected)
         for act in actual:
             assert act in expected
+
+
+@pytest.mark.parametrize(
+    ('var'),
```
spencerahill: Here and elsewhere: parentheses around a single object do nothing: `('var') == 'var'`.  So I would omit them, except where they are used to break expressions over multiple lines.

Conversely, if there is a trailing comma, it defines a length-1 tuple: e.g. `('var',)`.

aospy/test/test_automate.py
```diff
@@ -351,3 +354,25 @@ def test_permute_aux_specs(self, calc_suite, calcsuite_specs):
...
+    [(precip_largescale),
+     (condensation_rain)]
+)
+def test_prune_invalid_time_reductions(var):
+    time_options = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts', 'None']
+    spec = {
+        'var': var
+    }
+    for i in range(1, 8):
```
spencerahill: To be more explicit, replace `range(1, 8)` with `range(1, len(time_options) + 1)`.

spencerahill: You are using the string `'None'`, instead of the built-in `None` object.  I think we want the latter.

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,53 @@ def setUp(self):
...
+}
+
+
+@pytest.mark.parametrize(
+    ('dtype_out_time'),
+    [(None),
+     ('None'),
+     ([])]
+)
+def test_calc_object_no_time_options(dtype_out_time):
```
spencerahill: This test doesn't assert anything; it will only fail if the `CalcInterface` call fails.  So let's add an assertion about the `dtype_out_time` in the resulting `CalcInterface` object.

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,53 @@ def setUp(self):
...
+     ('reg.std'),
+     ('reg.ts')]
+)
+def test_calc_object_string_time_options(dtype_out_time):
+    test_params['dtype_out_time'] = dtype_out_time
+    with pytest.raises(ValueError):
+        CalcInterface(**test_params)
+
+
+@pytest.mark.fixture
```
spencerahill: The fixture decorator shouldn't be used on a test; instead it's used to prepare the data needed for a test.  See https://docs.pytest.org/en/latest/fixture.html#fixtures

docs/whats-new.rst
```diff
@@ -51,6 +51,9 @@ Bug Fixes
   writing to netCDF files to avoid bugs when using ``libnetcdf``
   version 4.5.0 (:pull:`235`).  By `Spencer Hill
   <https://github.com/spencerahill>`_.
+- Skip ``Calc`` object calculation that computes time average of
```
spencerahill: Expand a bit: note that this refers to when using `CalcSuite` (and thus `submit_mult_calcs`), and that `Calc` now raises a ValueError when instantiated with a non-time-defined variable but with one or more time-defined reductions.

spencerahill (changes requested):
Thanks @micahkim23 .  I think we're most of the way there.  A few things though that need fixing or I'm not totally understanding.

aospy/test/data/objects/examples.py
micahkim23: I just needed to test a `Var` object with `def_time=False`.

docs/whats-new.rst
```diff
@@ -51,9 +51,12 @@ Bug Fixes
   writing to netCDF files to avoid bugs when using ``libnetcdf``
   version 4.5.0 (:pull:`235`).  By `Spencer Hill
   <https://github.com/spencerahill>`_.
-- Skip ``Calc`` object calculation that computes time average of
-  a variable with no time dimension. (closes :issue:`202`
-  via :pull:`242`). By `Micah Kim <https://github.com/micahkim23>`_.
+- In ``CalcSuite`` when calling ``submit_mult_calc``, calculations
```
spencerahill: `CalcSuite` doesn't call `submit_mult_calc`; it's the other way around.

aospy/test/test_automate.py
```diff
@@ -351,3 +354,24 @@ def test_permute_aux_specs(self, calc_suite, calcsuite_specs):
         assert len(actual) == len(expected)
         for act in actual:
             assert act in expected
+
+
+@pytest.mark.parametrize(
```
spencerahill: Minor style note: splitting across lines like this, with an open parentheses/bracket/curly bracket and then an immediate line break, is usually only needed when either

1. The items/expression within the overall parentheses/bracket/curly bracket is long enough that this helps it fit onto fewer lines, or at least look nicer.
2. The items aren't themselves too long to fit on one line each, but there are a lot of them, so it makes it easier to read.

In this case, you don't need to split lines at all; it would fit into a single line:

```python
@pytest.mark.parametrize('var', [precip_largescale, condensation_rain])
```

Also note more meaningless parentheses; please omit.

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,56 @@ def setUp(self):
         }
 
 
+test_params = {
```
spencerahill, line 164: nit:  indentation is longer than the standard 4 spaces.

(C.f. comment above re: line-splitting: for me this particular case falls into a gray area of intermediate length, where it could work throwing everything onto as few lines as possible, but it's also long enough that what you've done also works.)

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,56 @@ def setUp(self):
...
+        'proj': example_proj,
+        'model': example_model,
+        'run': example_run,
+        'var': precip_largescale,
+        'date_range': 'default',
+        'intvl_out': 1,
+}
+
+
+@pytest.mark.parametrize(
```
spencerahill: More unneeded parentheses + line breaks

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,56 @@ def setUp(self):
...
+def test_calc_object_no_time_options(dtype_out_time):
+    test_params['dtype_out_time'] = dtype_out_time
+    calc = CalcInterface(**test_params)
+    if isinstance(dtype_out_time, list):
+        assert calc.dtype_out_time == tuple(dtype_out_time)
+    else:
+        assert calc.dtype_out_time == tuple([dtype_out_time])
+
+
+@pytest.mark.parametrize(
```
spencerahill, line 184: More unneeded parentheses

docs/whats-new.rst
```diff
@@ -51,6 +51,12 @@ Bug Fixes
   writing to netCDF files to avoid bugs when using ``libnetcdf``
   version 4.5.0 (:pull:`235`).  By `Spencer Hill
   <https://github.com/spencerahill>`_.
+- In ``CalcSuite`` when calling ``submit_mult_calc``, calculations
```
spencerahill: Other way around: `submit_mult_calc` calls `CalcSuite`.  How about "`CalcSuite` (and thus `submit_mult_calcs`) now skips calculations that involve..."

aospy/test/data/objects/examples.py
spencerahill: OK.  Let's rename it to make that explicit; otherwise it's confusing why there are two nearly identical objects.  E.g. `var_not_time_defined = Var(name='var_no_time_def', def_time=False)`.  You can leave `alt_names` and `description` unspecified.

spencerahill (changes requested):

aospy/automate.py
```diff
@@ -22,6 +22,7 @@
 _REGIONS_STR = 'regions'
 _VARIABLES_STR = 'variables'
 _TAG_ATTR_MODIFIERS = dict(all='', default='default_')
+_TIME_DEFINED_REDUCTIONS = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts']
```
spencerkclark: For now I think it might be a good idea to keep the definition of these only in `calc.py` and import them here.  This way we won't need to remember to maintain this list in two places. 

```python
from calc import _TIME_DEFINED_REDUCTIONS
```

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,45 @@ def setUp(self):
...
+    ('dtype_out_time'),
```
spencerkclark: I think you can remove the parentheses here too.

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,45 @@ def setUp(self):
...
+    'proj': example_proj,
+    'model': example_model,
+    'run': example_run,
+    'var': var_not_time_defined,
+    'date_range': 'default',
+    'intvl_out': 1,
+}
+
+
+@pytest.mark.parametrize(('dtype_out_time'), [None, 'None', []])
```
spencerkclark: I think you can remove the parentheses around `'dtype_out_time'` here.  Also, is there a reason we are testing with the value `'None'`?

aospy/test/test_calc_basic.py
```diff
@@ -160,6 +161,45 @@ def setUp(self):
...
+    ('dtype_out_time'),
+    ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts'])
+def test_calc_object_string_time_options(dtype_out_time):
+    test_params['dtype_out_time'] = dtype_out_time
+    with pytest.raises(ValueError):
+        CalcInterface(**test_params)
+
+
+def test_calc_object_time_options():
+    time_options = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts', 'None']
```
spencerkclark: Per @spencerahill's earlier comment, should we use `None` rather than `'None'` here?

spencerkclark (commented):
Almost there @micahkim23!  Just a few more small things from me.

aospy/test/test_calc_basic.py
micahkim23: https://github.com/spencerahill/aospy/blob/67ec6197a97a13dfe7a63ad7a555dcc9056bd8c8/aospy/calc.py#L486 @spencerkclark I thought `'None'` was a possible time reduction.

spencerkclark: Good catch @micahkim23!  @spencerahill was this intentional?  Should we change it?

spencerahill: Woops!  Yes, good call @micahkim23 .  Sorry if any comments I made previously were confused on this point.

Since `'None'` appears to just be a synonym for `'ts'`, I'd say we just remove it entirely.  So there's only `['ts', 'av', 'std', 'reg.ts', 'reg.av', 'reg.std']`.

Since custom reductions are a big priority moving forward (#208), this will get refactored soon-ish anyways.

spencerahill (approved):

spencerahill:
We tempted the AppVeyor gods too long 😜 

I'd say this is good to go.  Thanks @spencerkclark for the reviews and esp. @micahkim23 for the PR!    And sorry again for my leading you astray on the `None` business.  Nice to have this in.

Issue #249, opened 2017-12-17
Deprecate CalcInterface


spencerahill:
Similar in spirit to #248 , I think it would be useful for future refactoring efforts to continue pruning out obsolete code.  I believe that `CalcInterface` is in that category.

Basically, I implemented after learning about programming interfaces in a class.  But as far as I can tell, all `CalcInterface` does is add another layer of complexity.  I see no reason why e.g. `CalcSuite` shouldn't just call the `Calc` constructor directly, rather than creating the intermediate `CalcInterface` and then passing it to the `Calc` constructor.

But this one will be a little more involved than #248.  And I could be persuaded that we really do want it, or that it's not worth the time.  So I thought I'd open this Issue to elicit feedback.  I was thinking of doing this over the holidays...seems like the kind of low-mental-effort-required task for this time of year :)

spencerahill:
Closed by #250

PR #250, opened 2017-12-18
Deprecate CalcInterface

spencerahill:
C.f. #249.  Turns out this was easier than I thought it would be.

Although I accidentally forgot to pull the latest commits to my local develop branch before creating this branch, so I'll have to resolve the merge conflicts manually.

Given how easy this was, I'm inclined to proceed with it.

### DIFF
--- a/aospy/automate.py
+++ b/aospy/automate.py
@@ -10,7 +10,7 @@ import logging
 import pprint
 import traceback
 
-from .calc import Calc, CalcInterface
+from .calc import Calc, CalcInterface, _TIME_DEFINED_REDUCTIONS
 from .region import Region
 from .var import Var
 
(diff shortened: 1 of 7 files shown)


aospy/automate.py
```diff
@@ -237,7 +236,7 @@ def _combine_core_aux_specs(self):
 
     def create_calcs(self):
         """Generate a Calc object for each requested parameter combination."""
-        return [Calc(CalcInterface(**sp)) for sp in
+        return [Calc(**sp) for sp in
```
spencerkclark: Could this fit on one line now?

spencerkclark (approved):
Indeed this does seem pretty straightforward!  ~~It looks like you just need to merge with the latest master branch and you should be good.~~ 

Edit: obviously you're well aware of the need to merge and as you note above, of course it's the develop branch in our case :)

spencerahill:
Very weird, this test is passing on my local machine...

spencerahill:
Disregard previous comment; hadn't pushed latest changes.  In it goes!

PR #251, opened 2018-01-12
Update ahead of change of 'item in ds' behavior in xarray v0.11

spencerahill:
Just updated to xarray v0.10 and started getting a lot of deprecation warnings when running aospy due to https://github.com/pydata/xarray/pull/1645, c.f. https://github.com/pydata/xarray/issues/1267.  So I tried to find it everywhere this comes up and fix it.

@spencerkclark let me know if you see any reason why we shouldn't do this, or if I missed anything.

### DIFF
--- a/aospy/data_loader.py
+++ b/aospy/data_loader.py
@@ -176,7 +176,7 @@ def _prep_time_data(ds):
     """
     ds = times.ensure_time_as_dim(ds)
     ds, min_year, max_year = times.numpy_datetime_workaround_encode_cf(ds)
-    if TIME_BOUNDS_STR in ds:
+    if TIME_BOUNDS_STR in ds.coords:
         ds = times.ensure_time_avg_has_cf_metadata(ds)
         ds[TIME_STR] = times.average_time_bounds(ds)
     else:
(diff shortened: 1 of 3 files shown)


spencerkclark (commented):
@spencerahill thanks for making these updates.  It's clear we will need to be careful about this update to xarray.  

In these places it seems we were relying on `__contains__` working differently for DataArrays and Datasets.
- For DataArrays we were checking if a name was a key in the dictionary of coordinates.
- For Datasets we were checking if a name was a key in either the dictionary of coordinates or data variables.

Do you know where exactly the warnings were emanating from?  The warnings should only stem from when we use the `__contains__` operator on a DataArray.  If some of these methods are called using Datasets we probably want to check if the name is in the data variables dictionary as well (and therefore should use the original `VAR_STR in ds` logic).

My initial thought was that since `_prep_time_data` is called before `_sel_var` in `data_loader.load_variable`:
https://github.com/spencerahill/aospy/blob/389961acd046396b3b237b5b2749d70b345b1e6b/aospy/data_loader.py#L278-L280
 `ds` must always be a Dataset, but it seems if this warning was triggered, it must sometimes be a DataArray.  Did you encounter these warnings when doing calculations using aospy or when running the test suite?

spencerahill:
Oy, this is an embarrassing one.  I forgot that I was running on an old branch for one of my research projects.  I just ran the example main script and the test suite on develop and no warnings are emitted.

Thanks @spencerkclark for responding, sorry for wasting your time!

PR #254, opened 2018-03-21
Raise if submit_mult_calcs generates no calculations

### LINKED ISSUE #253: Warn (or perhaps raise?) when the main script is constructed to submit zero calculations
For instance if one forgets to specify `input_vertical_datatypes` and instead uses an empty list, the main script will run and silently compute nothing.  This is confusing for users who might not be familiar with the internals in `automate.py`, which create `Calc` objects based on the product of the input options (so if one option is an empty list, zero objects will be created).

spencerahill:
Closes #253 .

I elected to raise rather than warn, but I don't feel strongly either way.  @spencerkclark please give this a quick review when you have time.


spencerahill:
Travis failures are both kernel timeouts on our test of the tutorial notebook: https://travis-ci.org/spencerahill/aospy/jobs/356502217#L1082

These seem unrelated to this commit. 

Edit: both Travis and AppVeyor failures are the same 3.4 thing, and I've since merged #255 which gets rid of Python 3.4.  There's one Travis 2.7 failure that's a network/distributed thing, also unrelated (see #256)

aospy/automate.py
```diff
@@ -487,4 +487,10 @@ def submit_mult_calcs(calc_suite_specs, exec_options=None):
         _user_verify()
     calc_suite = CalcSuite(calc_suite_specs)
     calcs = calc_suite.create_calcs()
+    if len(calcs) == 0:
```
spencerkclark: nit: might `if not calcs:` be more idiomatic here?

spencerkclark (approved):
@spencerahill one minor nit for you to consider, but otherwise this looks great!

spencerahill:
In it goes!  Thanks @spencerkclark

PR #233, opened 2017-11-13
Fix AppVeyor 3.4 failing builds

spencerahill:
Closes #227 

My approach here is basically just guess-and-check until I find something that works, following solutions other packages have found from the link in #227 and within.

PR #255, opened 2018-03-21
Drop Python 3.4 support

spencerahill:
Necessary given https://github.com/pydata/xarray/issues/1829, and also will save us headaches from the recent spat of CI errors on 3.4 only (c.f. #254).


spencerahill:
The one Travis failure is just some network hangup on one of our tests with distributed.  Merging.

spencerkclark:
Thanks for doing this; the [previous issues we had with `scipy`](https://github.com/spencerahill/aospy/pull/233) in the test suite alone for Python 3.4 were enough to convince me :)

Issue #259, opened 2018-03-21
Travis intermittent failures on our submit_mult_calcs external client tests in 2.7 only


spencerahill:
This is starting to happen with more regularity now, e.g. https://travis-ci.org/spencerahill/aospy/jobs/356503697#L854.  I've pasted the full error message below.

Further motivation to drop 2.7? :stuck_out_tongue_winking_eye:  c.f. #256 

```
==================================== ERRORS ====================================
___ ERROR at setup of test_submit_mult_calcs_external_client[exec_options0] ____
    @pytest.fixture
    def external_client():
        # Explicitly specify we want only 4 workers so that when running on
        # Travis we don't request too many.
>       cluster = distributed.LocalCluster(n_workers=4)
aospy/test/test_automate.py:221: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../miniconda/envs/test_env/lib/python2.7/site-packages/distributed/deploy/local.py:127: in __init__
    self.start(ip=ip, n_workers=n_workers)
../../../miniconda/envs/test_env/lib/python2.7/site-packages/distributed/deploy/local.py:145: in start
    sync(self.loop, self._start, **kwargs)
../../../miniconda/envs/test_env/lib/python2.7/site-packages/distributed/utils.py:254: in sync
    six.reraise(*error[0])
../../../miniconda/envs/test_env/lib/python2.7/site-packages/distributed/utils.py:238: in f
    result[0] = yield make_coro()
../../../miniconda/envs/test_env/lib/python2.7/site-packages/tornado/gen.py:1099: in run
    value = future.result()
../../../miniconda/envs/test_env/lib/python2.7/site-packages/tornado/concurrent.py:260: in result
    raise_exc_info(self._exc_info)
../../../miniconda/envs/test_env/lib/python2.7/site-packages/tornado/gen.py:1107: in run
    yielded = self.gen.throw(*exc_info)
../../../miniconda/envs/test_env/lib/python2.7/site-packages/distributed/deploy/local.py:163: in _start
    yield [self._start_worker(**self.worker_kwargs) for i in range(n_workers)]
../../../miniconda/envs/test_env/lib/python2.7/site-packages/tornado/gen.py:1099: in run
    value = future.result()
[...]

PR #263, opened 2018-04-02
Recursively compute variables

### LINKED ISSUE #3: Var.vars should accept Vars that are themselves functions of multiple Vars

spencerkclark:
Closes #3 

This allows one to compose variable objects whose functions take other computed variables as arguments.  For example:
```python
from aospy import Var

a = Var(name='a')
b = Var(name='b', variables=(a, ), func=lambda x: 2. * x)
c = Var(name='c', variables=(a, b), func=lambda x, y: x + y)
```


spencerkclark:
Note of course this doesn't deal with saving intermediates to disk (or checking for saved intermediates), which is also discussed in #3, but I think what's here is already a reasonable improvement.

spencerahill:
Wow, this is surprisingly elegant!  The logic and tests look good to me.

One big thing: have we tested this out within a `Calc` yet?  I expect that, given these unit tests, everything will be fine, but I still think it would be good to have an explicit integration test for this.

Besides that, a couple documentation requests:

- Can you give `recursively_compute_variable` a full docstring?
- More substantively, can we highlight this functionality in the docs somewhere?  In scanning through our existing pages, I'm not sure where this fits best.  Perhaps [here](https://aospy.readthedocs.io/en/latest/examples.html#physical-variables) within the Examples page.

aospy/utils/io.py
```diff
@@ -154,9 +154,14 @@ def data_name_gfdl(name, domain, data_type, intvl_type, data_yr,
...
+
+    archive_files = []
+    for f in files_list:
+        if f.startswith('/archive'):
+            archive_files.append(f)
     try:
-        if isinstance(files_list, str):
-            files_list = [files_list]
-        subprocess.call(['dmget'] + files_list)
+        subprocess.call(['dmget'] + archive_files)
```
spencerkclark, line 165: I was finding that the test suite would hang if `dmget` was called on an existing file that was not stored in `/archive`.  I was able to reproduce this outside of aospy, so it has nothing to do with the changes I made to allow for the recursive calculation of variables; they must have made some updates to `dmget`.  The logic here filters out files that aren't stored on archive before passing them to `dmget`.

aospy/data_loader.py
```diff
@@ -296,7 +296,7 @@ def recursively_compute_variable(self, var, start_date=None, end_date=None,
             data = [self.recursively_compute_variable(
                 v, start_date, end_date, time_offset, **DataAttrs)
                     for v in var.variables]
-            return var.func(*data)
+            return var.func(*data).rename(var.name)
```
spencerkclark: Indeed, good suggestion to add a test for this.  One minor change and things work smoothly!

spencerkclark (commented):
Thanks @spencerahill!  I agree this should be highlighted in the documentation.  It will certainly simplify my object library quite a bit.  I'll add a full docstring to `recursively_compute_variable` and an example to the official docs sometime later today or tomorrow.

spencerahill:
All sounds/looks good.  Just ping me when you're ready for final review.

spencerkclark:
I guess I spoke too soon in https://github.com/spencerahill/aospy/issues/259; that's the failing test here :)

@spencerahill anyway, whenever you get a chance, this should be ready for a final review!

spencerahill:
>I guess I spoke too soon in #259; that's the failing test here :)

Bummer.  Still, it seems like the frequency of the failures has gone down...maybe it's on its way toward going away completely.

In it goes!  Thanks @spencerkclark, this is great.

PR #265, opened 2018-04-03
Delete _robust_bool and get_parent_attr

### LINKED ISSUE #264: utils.io._robust_bool and utils.io.get_parent_attr are no longer used
I was glancing at [a recent Coveralls build](https://coveralls.io/builds/16314863/source?filename=aospy%2Futils%2Fio.py) and noticed that we do not have test coverage for `utils.io._robust_bool` or `utils.io.get_parent_attr`.  A quick grep indicated that we do not use these functions internally within aospy anymore.  Is there a reason to keep them?

spencerkclark:
Closes #264

### DIFF
--- a/aospy/utils/io.py
+++ b/aospy/utils/io.py
@@ -5,36 +5,6 @@ import subprocess
 import numpy as np
 
 
-def _robust_bool(obj):
-    try:
-        return bool(obj)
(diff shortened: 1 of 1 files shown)


spencerahill:
Thanks!

PR #266, opened 2018-04-23
Region: explicit W/E/S/N bounds and cyclic longitude support; Longitude class and longitude module

### LINKED ISSUE #229: Region definitions assume [0, 360] longitudes range
Just noticed this while using `Region.ts` on data that is on a [-180, 180] longitude grid, using the following region:

```python
sahel = Region(
    name='sahel',
    description='African Sahel',
    mask_bounds=[((10, 20), (0, 40)), ((10, 20), (342, 360))],
    do_land_mask=True
)
```

The (342, 360) bit gets dropped, since it is out of the array's (-180, 180) range, causing the region to be truncated to (0, 40).  This is physically incorrect and thus a bug.

We need to translate all region definitions and the longitude arrays on which the regional reductions are operating to a standard longitude array, which by convention should be either (-180, 180) or (0, 360).  This can be accomplished by noting that `(0, 180) + 360*i` corresponds to the Eastern Hemisphere and `(180, 360) + 360*j` corresponds to the Western hemisphere, where `i` and `j` are integers.  E.g. `i=0, j=-1` corresponds to the (-180, 180) grid.

So the steps would be

1. Convert the region's longitude bounds to aospy's chosen reference longitude.
2. Convert the array's longitudes to aospy's chosen reference longitude.
3. Perform the reduction.

Once implemented, we should be able to support longitude arrays with arbitrary start and end values.

spencerahill:
Closes #229 

Still need to implement the cyclic longitudes logic we converged on in #229; the new logic that's currently in `region.py` that touches this will be updated, as will the related tests.

Still also need tests and to finish implementation of support for non-aospy data.  But wanted to open this PR to get a stake in the ground.

@spencerkclark running the tests on my local machine, the tutorial notebook test failed because I didn't update the `Region` definitions in the notebook to the new call signature.  But now I can't remember how we generated the tutorial notebook.  Do you recall?  I could manipulate the .ipynb file directly, but this seems fraught.


spencerkclark:
> But now I can't remember how we generated the tutorial notebook. Do you recall? I could manipulate the .ipynb file directly, but this seems fraught.

Yeah don't edit the the .ipynb file directly :). The easiest thing is to just open it in Jupyter, make the changes, and save it.

spencerahill:
The `calc` test suite is now failing across the board due to a problem with the tearDown, e.g. [here](https://travis-ci.org/spencerahill/aospy/jobs/370355983).

I don't see an obvious way that what I introduced caused this, although this is the first time we're seeing it.  I just added a try/except guard to see if that solves it.

spencerahill:
>I just added a try/except guard to see if that solves it.

Hmm, now these have been replaced by even less intuitive ones, e.g. [here](https://travis-ci.org/spencerahill/aospy/jobs/370365778)

It seems like the example objects within the CI environments have somehow been corrupted.  Maybe due to the change in Region call signature?  In contrast, these tests are all passing on my local machine.

spencerkclark:
> Hmm, now these have been replaced by even less intuitive ones, e.g. here

I think those were always there -- see L1828-L1873 in the [old build](https://travis-ci.org/spencerahill/aospy/jobs/370355983#L1828-L1873) for example.  The tear down errors were just side effects of the results never being computed/saved to files.

Xarray recently released a new version (version 0.10.3); which version are you on locally?

spencerahill:
>Xarray recently released a new version (version 0.10.3); which version are you on locally?

Good call.  I was on 10.2; after updating to 10.3 I can reproduce.  I'll open a separate issue to track this.

aospy/region.py
```diff
@@ -109,14 +132,15 @@ def __init__(self, name='', description='', lon_bounds=[], lat_bounds=[],
         description : str, optional
              A description of the region.  This is not used internally by
              aospy; it is solely for the user's information.
-        lon_bounds, lat_bounds : length-2 sequence, optional
-             The longitude and latitude bounds of the region.  If the region
-             boundaries are more complicated than a single lat-lon rectangle,
-             use `mask_bounds` instead.
+        west_bound, east_bound, south_bound, north_bound : scalar, optional
+             The western, eastern southern, and northern boundaries,
```
spencerkclark: Typo: add a comma between "eastern" and "southern."

aospy/region.py
```diff
@@ -92,14 +104,25 @@ class Region(object):
...
+    def __init__(self, name='', description='', west_bound=None,
+                 east_bound=None, south_bound=None, north_bound=None,
+                 mask_bounds=None, do_land_mask=False):
         """Instantiate a Region object.
 
-        If a region spans across the endpoint of the data's longitude array
-        (i.e.  it crosses the Prime Meridian for data with longitudes spanning
-        0 to 360), it must be defined as the union of two sections extending to
-        the east and to the west of the Prime Meridian.
+        Note that longitudes spanning (-180, 180), (0, 360), or any other range
```
spencerkclark, line 99: Should `(0, 360)` be `(180, 540)` to correspond with the example after the colon?

aospy/region.py
```diff
@@ -130,63 +154,174 @@ def __init__(self, name='', description='', lon_bounds=[], lat_bounds=[],
 
         Examples
         --------
-        Define a region spanning the entire globe
+        Define a region spanning the entire globe:
 
-        >>> globe = Region(name='globe', lat_bounds=(-90, 90),
-        ...                lon_bounds=(0, 360), do_land_mask=False)
+        >>> globe = Region(name='globe', west_bound=0, east_bound=360,
+        ...                south_bound=0, north_bound=360, do_land_mask=False)
```
spencerkclark: I'm guessing for this example `south_bound=-90` and `north_bound=90` should be used.

spencerkclark (commented):
I still need to look at the code, but I read through the docstrings (which are in general very nice!) and noticed a few minor things that I didn't want to forget.

spencerahill:
>I still need to look at the code, but I read through the docstrings (which are in general very nice!) and noticed a few minor things that I didn't want to forget.

Thanks!  FYI I've realized after this commit that the logic I implemented is incomplete (i.e. wrong)...I'm working on a better solution and will ping you when it's ready for a proper review.

spencerahill:
OK, c.f. #272, I have re-done the logic using the new utils.longitude module and Longitude class.  Remaining items to address for my own organizational purposes:

- [x] Make a decision about lon_cyclic ~(actually I think I've worked this out; new commit forthcoming)~
- [x] Add Longitude to main docs
- [x] Double-check docs for Region
- [x] What's new

spencerahill:
C.f. my comment https://github.com/spencerahill/aospy/issues/267#issuecomment-385822091, I think after all that dealing with non-cyclic longitudes is probably outside of the scope of this PR (despite that being partly the original motivation).  Frankly, I am almost out of spare cycles for the next few weeks, and I want to get what's here so far in.

So we would return to the cyclic points in conjunction with how we treat region boundaries in a future PR.  @spencerkclark does that sound OK to you?  Sorry that I turned this one into such a long and winding road.

Either way, I'd appreciate a review of this PR when you get the chance.  Thanks!

spencerkclark:
> So we would return to the cyclic points in conjunction with how we treat region boundaries in a future PR. @spencerkclark does that sound OK to you?

Totally fine!  I'll try and give things a review over the next few days.

aospy/utils/longitude.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def __repr__(self):
+        return "Longitude('{0}{1}')".format(self.longitude, self.hemisphere)
+
+    @other_to_lon
+    def __eq__(self, other):
+        if isinstance(other, Longitude):
+            cond = (self.hemisphere == other.hemisphere and
+                    self.longitude == other.longitude)
+            if cond:
```
spencerkclark: I believe this `if`-`else` block is equivalent to simply returning `cond`.

aospy/test/test_utils_longitude.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@pytest.mark.parametrize(
+    ('obj1', 'obj2'),
+    [(Longitude('100W'), Longitude('100W')),
+     (Longitude('90E'), Longitude('90E')),
+     (Longitude(0), Longitude(0)),
+     (Longitude('0E'), 0),
+     (Longitude('0E'), 720),
+     (Longitude('0E'), [0, 720])])
```
spencerkclark: Here and for the other operators, should we test arrays on either the right or left side of the operator?  For example:
```python
@pytest.mark.parametrize(
('obj1', 'obj2'),
[(Longitude('100W'), Longitude('100W')),
...
(Longitude('0E'), [0, 720]),
([0, 720], Longitude('0E'))])
```

aospy/utils/longitude.py
```diff
@@ -0,0 +1,210 @@
...
+                return True
+            else:
+                return False
+        else:
+            return np.equal(other, self)
```
spencerkclark: Here and in the other operators we need special logic if `other` is a DataArray; as written this currently doesn't work:
```
In [1]: import xarray as xr

In [2]: from longitude import Longitude

In [3]: Longitude(0.) == xr.DataArray([0., 720])
<ipython-input-3-85150267fe77> in <module>()
----> 1 Longitude(0.) == xr.DataArray([0., 720])

/Users/spencerclark/Downloads/lon_sah2/longitude.py in func_other_to_lon(self, other)
     72 def other_to_lon(func):
     73     def func_other_to_lon(self, other):
---> 74         return func(self, _maybe_cast_to_lon(other))
     75     return func_other_to_lon
     76

/Users/spencerclark/Downloads/lon_sah2/longitude.py in __eq__(self, other)
    136                 return False
    137         else:
--> 138             return np.equal(other, self)
    139
    140     @other_to_lon

TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'equal'>, '__call__', <xarray.DataArray (dim_0: 2)>
array([  0., 720.])
Dimensions without coordinates: dim_0, Longitude('0.0E')): 'DataArray', 'Longitude'
```
I think the following should fix things:
```python
if isinstance(other, xr.DataArray):
    return xr.apply_ufunc(np.equal, other, self)
else:
    return np.equal(other, self)
```

aospy/test/test_utils_longitude.py
spencerkclark: It would also be good to add test coverage for use with DataArrays.

aospy/region.py
```diff
@@ -130,63 +155,198 @@ def __init__(self, name='', description='', lon_bounds=[], lat_bounds=[],
...
+                else:
+                    lons = _lon_bounds_to_lon_obj(rect_bounds[:2])
+                    bounds.append(tuple(lons + list(rect_bounds[2:])))
+            self.mask_bounds = tuple(bounds)
+
     def __str__(self):
         return 'Geographical region "' + self.name + '"'
 
-    __repr__ = __str__
+    def _make_mask(self, data, lon_str=LON_STR, lat_str=LAT_STR):
```
spencerkclark, line 219: What is the reasoning behind having this function take `lon_str` and `lat_str` as arguments?  As part of the data-loading process don't we automatically rename the longitude and latitude coordinates `LON_STR` and `LAT_STR`?  Is it that you are envisioning using these functions outside of `aospy`?

aospy/region.py
```diff
@@ -130,63 +155,198 @@ def __init__(self, name='', description='', lon_bounds=[], lat_bounds=[],
...
+        else:
+            bounds = []
+            for rect_bounds in mask_bounds:
+                if len(rect_bounds) != 4:
+                    raise ValueError("Each element of `mask_bounds` must be a "
+                                     "length-4 array with values (west, east, "
+                                     "south, north).  Value given: "
+                                     "{}".format(rect_bounds))
+                else:
+                    lons = _lon_bounds_to_lon_obj(rect_bounds[:2])
```
spencerkclark: Even though it's an extra line, I have a slight preference for using tuple unpacking to extract the bounds here before converting to Longitude objects (for clarity):
```python
west_bound, east_bound, south_bound, north_bound = rect_bounds
lons = _lon_bounds_to_lon_obj([west_bound, east_bound])
bounds.append(tuple(lons + [south_bound, north_bound]))
```

aospy/test/test_region.py
```diff
@@ -23,62 +38,185 @@ def data_for_reg_calcs():
...
+                  (Longitude(0.), Longitude(5), 0, 90.))
+    assert region.do_land_mask is True
+
+
+def test_region_init_mult_rect():
+    bounds_in = [[1, 2, 3, 4], (-12, -30, 2.3, 9)]
+    region = Region(name='test', mask_bounds=bounds_in)
+    assert isinstance(region.mask_bounds, tuple)
+    assert len(region.mask_bounds) == 2
+    for n, bounds in enumerate(region.mask_bounds):
```
spencerkclark: Slight preference for using `zip` and tuple unpacking here rather than `enumerate` and indexing the `bounds_in` list/tuples:
```python
for (west, east, south, north), bounds in zip(bounds_in, region.mask_bounds):
    assert isinstance(bounds, tuple)
    assert np.all(bounds == (Longitude(west), Longitude(east), south, north))
```

aospy/test/test_region.py
```diff
@@ -23,62 +38,185 @@ def data_for_reg_calcs():
...
+
+
+def test_region_init_bad_bounds():
+    with pytest.raises(ValueError):
+        Region(mask_bounds=[(1, 2, 3)])
+        Region(mask_bounds=[(1, 2, 3, 4),
+                            (1, 2, 3)])
+
+
+# def test_make_mask_latlon_rect(data_for_reg_calcs):
```
spencerkclark: Is there a reason behind these commented-out tests?

spencerkclark (commented):
Thanks @spencerahill, this is going to be a big upgrade for our region handling.  I have mostly cosmetic comments, the one exception being one on the comparison of `Longitude` objects to DataArrays.

aospy/region.py
spencerahill: That's right.   This gets called by `mask_var`, which in turn gets called by `ts` and its siblings, which I want to make available to non-aospy data.

aospy/region.py
spencerahill: I'm going to switch to a namedtuple, which will make this logic cleaner

aospy/test/test_region.py
spencerahill: Sorry, forgot to remove these

spencerahill:
@spencerkclark thanks, your comments all look good.  Will address + flesh out the docs.

spencerahill:
OK, I think this is ready to go.  Will merge once tests pass (EDIT and add a What's New).  Thanks @spencerkclark for the invaluable help.

PR #269, opened 2018-04-25
Workaround for bug w/ len-1 times

### LINKED ISSUE #268: Failing test_calc_basic tests after updating xarray from 10.2 to 10.3
Originally noted [here](https://github.com/spencerahill/aospy/pull/266#issuecomment-383776179)

All of the failures are the same error.  Note that this one doesn't involve regional averaging, so it's unrelated to what I'm implementing in #266.  Here's the traceback of one:

```
$ py.test test_calc_basic.py::TestCalc3D::test_monthly_ts
================================================================== test session starts ==================================================================
platform darwin -- Python 3.6.3, pytest-3.2.3, py-1.5.1, pluggy-0.4.0
rootdir: /Users/shill/Dropbox/py/aospy, inifile: setup.cfg
plugins: catchlog-1.2.2, hypothesis-3.50.2
collected 1 item

test_calc_basic.py F

======================================================================= FAILURES ========================================================================
______________________________________________________________ TestCalc3D.test_monthly_ts _______________________________________________________________

self = <aospy.test.test_calc_basic.TestCalc3D testMethod=test_monthly_ts>

    def test_monthly_ts(self):
        calc = Calc(intvl_out=1, dtype_out_time='ts', **self.test_params)
>       calc.compute()

calc       = <aospy.Calc instance: sphum, example_proj, example_model, example_run>
self       = <aospy.test.test_calc_basic.TestCalc3D testMethod=test_monthly_ts>

test_calc_basic.py:88:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../calc.py:569: in compute
    self.end_date),
../calc.py:415: in _get_all_data
    for n, var in enumerate(self.variables)]
../calc.py:415: in <listcomp>
    for n, var in enumerate(self.variables)]
../calc.py:367: in _get_input_data
    **self.data_loader_attrs)
../data_loader.py:278: in load_variable
    ds, min_year, max_year = _prep_time_data(ds)
../data_loader.py:180: in _prep_time_data
    ds = times.ensure_time_avg_has_cf_metadata(ds)
../utils/times.py:417: in ensure_time_avg_has_cf_metadata
    raw_start_date = ds[TIME_BOUNDS_STR].isel(**{TIME_STR: 0, BOUNDS_STR: 0})
../../../../miniconda3/envs/py36/lib/python3.6/site-packages/xarray/core/dataarray.py:754: in isel
    ds = self._to_temp_dataset().isel(drop=drop, **indexers)
../../../../miniconda3/envs/py36/lib/python3.6/site-packages/xarray/core/dataset.py:1391: in isel
    indexers_list = self._validate_indexers(indexers)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <xarray.Dataset>
Dimensions:       (bounds: 2)
Coordinates:
    time_bounds   (bounds) float64 dask.array<shape=(2,), ...    time_weights  float64 ...
Data variables:
    <this-array>  (bounds) float64 dask.array<shape=(2,), chunksize=(2,)>
indexers = {'bounds': 0, 'time': 0}

    def _validate_indexers(self, indexers):
        """ Here we make sure
            + indexer has a valid keys
            + indexer is in a valid data type
            """
        from .dataarray import DataArray

        invalid = [k for k in indexers if k not in self.dims]
        if invalid:
>           raise ValueError("dimensions %r do not exist" % invalid)
E           ValueError: dimensions ['time'] do not exist

DataArray  = <class 'xarray.core.dataarray.DataArray'>
indexers   = {'bounds': 0, 'time': 0}
invalid    = ['time']
self       = <xarray.Dataset>
Dimensions:       (bounds: 2)
Coordinates:
    time_bounds   (bounds) float64 dask.array<shape=(2,), ...    time_weights  float64 ...
Data variables:
    <this-array>  (bounds) float64 dask.array<shape=(2,), chunksize=(2,)>
```

spencerahill:
Closes #268 

What's happening is that, with xarray 10.3, if time is length-1, then `time_bounds`, `time_weights`, and any data variables aren't being indexed by it.  I implemented the hack below that seems to fix it.

However, this feels really clunky, and I suspect there's a much more elegant way of doing it.  @spencerkclark, do you have any ideas?  The frustrating part was that I couldn't just use `DataArray.assign_coords`, because of a "ValueError: cannot add coordinates with new dimensions to a DataArray"

I'll wait to add regression tests etc. until we converge on the overall logic.


spencerkclark:
@spencerahill thanks for looking into this!  Following up on my comment in #268, I think the solution is to change `times.ensure_time_as_dim` to the following:
```python
def ensure_time_as_dim(ds):
    TIME_INDEXED_VARS = set(ds.data_vars).union({TIME_WEIGHTS_STR, TIME_BOUNDS_STR})
    TIME_INDEXED_VARS = TIME_INDEXED_VARS.intersection(ds.variables)
    for name in TIME_INDEXED_VARS:
        if TIME_STR not in ds[name].dims:
            da = ds[name].expand_dims(TIME_STR)
            da[TIME_STR] = ds[TIME_STR]
            ds[name] = da
    return ds
```
If we use `expand_dims` here we can do away with `times.convert_scalar_to_indexable_coord`.  I tested this locally, and it seems to do the trick.

spencerkclark:
Note when testing above I only tested `test_calc_basic.py`.  It seems the above implementation breaks `test_utils_times.test_ensure_time_as_dim`.  There is a simple fix which is to do the following:
```python
def ensure_time_as_dim(ds):
    if TIME_STR not in ds[TIME_STR].dims:
        ds[TIME_STR] = ds[TIME_STR].expand_dims(TIME_STR)

    time_indexed_vars = set(ds.data_vars).union(set(TIME_VAR_STRS))
    time_indexed_vars = time_indexed_vars.intersection(ds.variables)
    for name in time_indexed_vars:
        if TIME_STR not in ds[name].dims:
            da = ds[name].expand_dims(TIME_STR)
            da[TIME_STR] = ds[TIME_STR]
            ds[name] = da
    return ds
```
That said, we could consider modifying the test, because I think as of Xarray 0.10.3 we are guaranteed that `TIME_STR` will be a dimension for `ds[TIME_STR]`, since we call `xr.open_mfdataset` with `concat_dim=TIME_STR`.

spencerahill:
Thanks @spencerkclark, that was super helpful.  I slightly modified your last version, and now all tests pass.  Let me know your take.

>as of Xarray 0.10.3 we are guaranteed that TIME_STR will be a dimension for ds[TIME_STR], since we call xr.open_mfdataset with concat_dim=TIME_STR.

This is a good point.  So what I have implemented may be slightly overkill.  But I'm not sure it's worth the effort to spend much more time to prune this down.  For one, any performance hit will likely be small, since by definition this arises only if there is a single time value.

spencerahill:
Note that the failures above are all from #259; the actual relevant tests are passing

aospy/utils/times.py
```diff
@@ -559,17 +559,16 @@ def ensure_time_as_dim(ds):
...
-                da, _ = xr.broadcast(da, time)
-                da[TIME_STR] = time
-                ds[name] = da
+    time_indexed_vars = set(ds.data_vars).union({TIME_WEIGHTS_STR,
+                                                 TIME_BOUNDS_STR})
+    time_indexed_vars = time_indexed_vars.intersection(ds.variables)
+    for name in time_indexed_vars:
+        if TIME_STR not in ds[name].indexes:
+            da = ds[name].expand_dims(TIME_STR)
+            da[TIME_STR] = convert_scalar_to_indexable_coord(ds[TIME_STR])
```
spencerkclark: I'm slightly partial to a solution that eliminates `convert_scalar_to_indexable_coord` for the simple reason that it is something that Xarray supports now (so we shouldn't need to roll our own solution)

spencerkclark (commented):
> But I'm not sure it's worth the effort to spend much more time to prune this down.

I agree, not worth the effort.

I have one minor comment on preference, but overall this looks good to me.  Thanks!

spencerahill:
@spencerkclark good call.  I got rid of that function, renamed the other one, and revamped the tests to more directly check what we're interested in.  I think it's good to go, but I'd appreciate one last review.

aospy/utils/times.py
```diff
@@ -562,13 +562,16 @@ def ensure_time_as_index(ds):
...
     time_indexed_vars = time_indexed_vars.intersection(ds.variables)
     for name in time_indexed_vars:
         if TIME_STR not in ds[name].indexes:
             da = ds[name].expand_dims(TIME_STR)
             da[TIME_STR] = ds[TIME_STR]
             ds[name] = da
     # Convert time weights and bounds back to coordinates from data_vars
-    return ds.set_coords([TIME_WEIGHTS_STR, TIME_BOUNDS_STR])
+    for coord in time_indexed_coords:
+        if coord in ds.coords:
```
spencerkclark: Should this be `if coord not in ds.coords`?  Otherwise this loop is a no-op right?  

More generally though, I'm not sure if this logic is needed, because we call `set_grid_attrs_as_coords` directly after `_prep_time_data`, and all the functions called within `_prep_time_data` operate on Datasets:
https://github.com/spencerahill/aospy/blob/7e3e5becb6920c3c5987a6dfd0a398c85d918ae4/aospy/data_loader.py#L278-L279

spencerkclark: > Should this be `if coord not in ds.coords`?  Otherwise this loop is a no-op right?  

To cover the case where `TIME_WEIGHTS_STR` or `TIME_BOUNDS_STR` are not in the Dataset, I mean:  `if coord in ds.data_vars`

spencerahill: Good call.  For some reason while putting the tests together, I convinced myself that time_bounds and time_weights were being promoted to data_vars.  But the full suite passed on my machine with this omitted, so clearly that was mistaken.

spencerkclark (approved):
Great, I say go ahead and merge once the CI comes back green!

spencerahill:
Thanks for the invaluable help as always @spencerkclark

PR #275, opened 2018-06-05
Uncomment string indexing with length-1 CFTimeIndex test

spencerkclark:
I added a bug fix in xarray that made it into the latest release: https://github.com/pydata/xarray/pull/2166.  I just wanted to make this update before I forgot!


spencerahill:
Thanks!

PR #277, opened 2018-06-09
Improve GFDLDataLoader IOError message

spencerkclark:
Currently if files are not found for a variable using the `GFDLDataLoader`, the error message looks like:
```
OSError: Files for the var Var instance "example" cannot be locatedusing GFDL post-processing conventions
```
In addition to the lack of a space between "located" and "using," this is not very helpful for debugging, because it does not help one figure out what is wrong with the path names it tries.  This PR fixes that by printing out the attempted file names in the error message, e.g.:
```
OSError: Files for the var Var instance "example" cannot be located using GFDL post-processing conventions. Attempted using the following sets of paths:

[['/archive/Spencer.Clark/fre_stem/case/gfdl.ncrc3-default-repro/pp/atmos_native/ts/6hr/1yr/atmos_native.0010010100-0010123123.example.nc',
  '/archive/Spencer.Clark/fre_stem/case/gfdl.ncrc3-default-repro/pp/atmos_native/ts/6hr/1yr/atmos_native.0011010100-0011123123.example.nc',
  '/archive/Spencer.Clark/fre_stem/case/gfdl.ncrc3-default-repro/pp/atmos_native/ts/6hr/1yr/atmos_native.0012010100-0012123123.example.nc']]
```


spencerahill:
Awesome.  Thanks!  I think this is small enough that we can skip the What's New.  So I'll just merge now.

PR #278, opened 2018-06-11
Remove obsolete eddy functionality

### LINKED ISSUE #274: Decide what to do with eddy-related functionality in Calc
2.5 years ago I implemented support for "eddy" fields, meaning the anomalies of a field with respect to it's monthly mean.  And likewise for the zonally asymmetric component of fields, i.e. the difference between a field's full value and its zonal mean value.  All of the related code is in `Calc`.

This code is quite convoluted and hackish.  It also is not even documented at all in our official docs, nor can I remember the last time I used it...probably not since right after I implemented it!  It's also now complicating our efforts to switch to the newly available CFTimeIndex (see https://github.com/spencerahill/aospy/pull/273#issue-189192781).  So we should decide what to do with it.

In my opinion, the best course of action would be to simply strip it all out for now and then re-implement it in a non-cringe-worthy way if/when the use case arises.  But if others (@spencerkclark, @chuaxr) are using it, we could leave it for now.

spencerahill:
Closes #274 

In starting work on #208, I was getting confused by the eddy stuff.  It turned out to be pretty easy to remove.  The Calc internals remain painfully convoluted even with this out, but this is a small step in the right direction.

@spencerkclark does this look OK to you?


spencerkclark (approved):
Great, yes this looks good to me.  Thanks!

spencerahill:
Thanks!

PR #279, opened 2018-06-11
Remove obsolete pytest-catchlog requirement

### LINKED ISSUE #270: Use builtin catchlog features in new pytest versions
Now getting the following warning on our tests:

```
===================================================================== warnings summary ======================================================================
None
  pytest-catchlog plugin has been merged into the core, please remove it from your requirements.

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

We should take care of this at some point.

spencerahill:
Closes #270 

Also added a `catch_warnings` to the tutorial notebook test so that we stop getting meaningless warnings about a "sticky bit".

With these in place, our test suite should now pass (at least on my machine it does) with zero warnings.  And let's commit to maintaining that moving forward, so that when warnings do arise we can be sure that they're relevant.


spencerkclark (approved):
> And let's commit to maintaining that moving forward, so that when warnings do arise we can be sure that they're relevant.

Agreed!

aospy/test/test_automate.py
```diff
@@ -219,7 +219,7 @@ def external_client():
     cluster = distributed.LocalCluster(n_workers=4)
     client = distributed.Client(cluster)
     yield client
-    client.shutdown()
+    client.close()
```
spencerkclark, line 222: Ah good catch!

spencerahill:
Thanks!

PR #280, opened 2018-06-11
Drop support for Python 2.7

### LINKED ISSUE #256: Drop Python 2.7 support
Xarray is planning on doing so by the end of 2018, c.f. https://github.com/pydata/xarray/issues/1830

So we'll need to do it no later than they do.  I'm fine with doing it earlier than that, provided it doesn't cause undue problems for our  users.  Are any of our current users using 2.7 and would be hard pressed to move to 3.x?  @chuaxr what are you using?

At the same time, I don't have any immediately pressing need to switch to only 3.x compatibility, so there's no need to rush.  @spencerkclark , do you have any ideas of what this could open up for us?  The only thing for me is if we one day get into type annotations, c.f. #81 , since the syntax is much cleaner for 3.x.  But that's very low priority.

A more immediate benefit would be simply one less CI environment to manage, which isn't trivial.

Related, we are dropping 3.4 shortly: #255

spencerahill:
Closes #256 

We don't have any known users on 2.7, so I suspect we'd be fine just going ahead with this before the 0.3 release.  But I'm not in a hurry to merge; just thought I'd put this up this while I was doing a few other cleanup PRs


spencerkclark (commented):
Sounds good; I agree we might as well just go through with it now.  At the very least it will make the CI faster!

I think you may have inadvertently added some lock files from the dask-worker-space to this commit.

spencerahill:
>At the very least it will make the CI faster!

Ha yes my thoughts exactly.

OK, I'll go ahead with a merge then once we're sure I've got the CI configuration etc. all working and we're green.

>I think you may have inadvertently added some lock files from the dask-worker-space to this commit.

Oops; good catch.

spencerkclark:
Looks good to me; thanks @spencerahill!

spencerahill:
Great, will go ahead with this then.

FYI I reverted the `install_python.ps1` file because I was getting errors from AppVeyor like [this one](https://ci.appveyor.com/project/spencerahill/aospy/build/1.0.560).  But I also turned off the `build` setting within the AppVeyor account, which I actually suspect is what caused it to start working again.  But it's not worth it to try re-modifying the install script...it works so let's move on.

PR #283, opened 2018-06-12
Update year in README

spencerkclark:
In the spirit of the doc updates in #282, we should also update the year in the README.


spencerahill:
Thanks!

PR #284, opened 2018-06-12
Remove obsolate subsetting in Calc.load

spencerahill:
Another in the vein of #282, this one for data subsetting within Calc's loading functionality, none of which was documented and most of which was just placeholders for future implementations.

Note that I do use `Calc.load` itself, still, so that should remain for now.  Although I think it would still make sense to segregate it from `Calc` itself eventually...maybe a `CalcDataLoader`?  @spencerkclark what do you think?  We can open a separate issue to track that.


spencerkclark (approved):
> Note that I do use Calc.load itself, still, so that should remain for now. Although I think it would still make sense to segregate it from Calc itself eventually...maybe a CalcDataLoader?

Yes I think there is a lot of room for improvement here.  When plotting data produced by aospy it is often helpful to not just load individual calculations, but to load groups of them.  For example something as simple as loading a particular calculation for a set of runs, and concatenating along a new dimension indexed by the run names can be very convenient.  I've written various functions over the years to do variants of that where needed.

Issue #170, opened 2017-04-03
Whither making core objects hashable


spencerahill:
Copying relevant discussion from #111, which merits further discussion here:

>Note that converting a list of Model objects to a set relies on aospy core objects being hashable, which they currently are not

>This is a good point. IIRC you dealt with hashes some when experimenting with a DB implementation. What’s your take, is it reasonable to consider Proj/Model/Runs immutable & hashable (and therefore fair game to implement hash)? My intuition is yes: I don’t see how the core attributes that differentiate them from others of their same type would change during their lifetimes.

>But, now that I think about it, does this require preventing setting of those core attributes after instantiation (i.e. by making them properties and using @property.setter). Consider model1 = Model(‘model’, runs=None, …) followed later by model1.runs = [run1, run2]. Shouldn’t that change the hash? As a concrete example, the Examples page does this with proj.regions.

>Indeed, I think we need to think more carefully about this. Typically mutable collections are not hashable; therefore, in the strict sense I think perhaps we'd have to think against making Proj and Model objects hashable.

PR #287, opened 2018-06-14
Refactor pressure handling in calc.py

spencerkclark:
@spencerahill in the spirit of cleaning up Calc, I figured I would take a shot at refactoring the pressure logic.  This is not completely addressing #80; however, I do feel it is an improvement from what we currently do.

Here I make pressure variables first-class `Var` objects and load data for them just as we would for any other variable.  This required a couple of changes:
- Adding an optional `model` argument to `data_loader.load_variable`; this allows one to specify the locations of grid-attributes for a particular run in a `DataLoader` or fall back on a `Model` object for loading them (before we only loaded grid attributes specified in `Var` function call signatures from `Model` objects)
- Making our logic for interpreting `Var` objects in `Var` function call signatures with names `'p'` and `'dp'` less opaque; here I explicitly replace those `Var` objects with the appropriate `Var` from `utils.vertcoord` depending on the `dtype_in_vert`.

I think I should add some explicit tests for passing a `Model` object to `data_loader.load_variable`, but this is done implicitly in the existing test suite, and things are passing.  This would also be a good time to document how to use pressure variables in one's own computed `Var` objects.

Before I do that though, do you agree with this approach?


aospy/calc.py
```diff
@@ -13,35 +13,32 @@
...
-    def_lon=True,
-)
+_P_VARS = {internal_names.ETA_STR: utils.vertcoord.p_eta,
+           internal_names.PLEVEL_STR: utils.vertcoord.p_level}
+_DP_VARS = {internal_names.ETA_STR: utils.vertcoord.dp_eta,
+            internal_names.PLEVEL_STR: utils.vertcoord.dp_level}
 _TIME_DEFINED_REDUCTIONS = ['av', 'std', 'ts', 'reg.av', 'reg.std', 'reg.ts']
 
 
+def _replace_pressure(variables, dtype_in_vert):
```
spencerahill: Maybe `vars_out` would be more readable than `_variables`

aospy/utils/vertcoord.py
```diff
@@ -262,3 +263,46 @@ def does_coord_increase_w_index(arr):
         raise ValueError("Array is not monotonic: {}".format(arr))
     # Since we know its monotonic, just test the first value.
     return bool(diff[0])
+
+
+bk = Var(
```
spencerahill, line 268: For a series of object definitions like this, I think the most readable is to have every kwarg on its own line.  Even though it ends up taking quite a few more lines.

aospy/data_loader.py
```diff
@@ -270,25 +272,37 @@ def load_variable(self, var=None, start_date=None, end_date=None,
...
+            )
+            if var.def_time:
+                ds = _prep_time_data(ds)
+                start_date = times.maybe_convert_to_index_date_type(
+                    ds.indexes[TIME_STR], start_date)
+                end_date = times.maybe_convert_to_index_date_type(
+                    ds.indexes[TIME_STR], end_date)
+            ds = set_grid_attrs_as_coords(ds)
+            da = _sel_var(ds, var, self.upcast_float32)
+            if var.def_time:
```
spencerahill: I suspect that these two separate `if var.def_time:` blocks are in fact necessary, otherwise you would have combined them :)

aospy/data_loader.py
```diff
@@ -270,25 +272,37 @@ def load_variable(self, var=None, start_date=None, end_date=None,
...
-        da = self._maybe_apply_time_shift(da, time_offset, **DataAttrs)
-        return times.sel_time(da, start_date, end_date).load()
+        try:
+            file_set = self._generate_file_set(var=var, start_date=start_date,
+                                               end_date=end_date, **DataAttrs)
+        except (IOError, KeyError) as e:
+            if var.name not in GRID_ATTRS or model is None:
+                raise e
+            else:
+                return getattr(model, var.name)
```
spencerahill: This will raise a generic AttributeError if the model doesn't have the attribute.  I think it's worth adding our own error message, providing the context that this is within the data loader.

spencerahill (commented):
Thanks @spencerkclark for taking this on.  It really cleans up `calc.py` quite a bit.

My only real concern (other comments are minor/style): I definitely see the expediency of adding the logic for getting data from a model to `DataLoader.load_variable`.  But it feels a bit out of place there, like we're shimming it in.

What about a separate utility function, something like `_load_or_get_from_model`, that encapsulates the try/except clause.  I.e. (pseudocode)

```python
def _load_or_get_from_model(...):
    try:
        data_loader.load_variable(...)
    except (...):
        if var.name not ...
        ...
```

What do you think?

aospy/data_loader.py
spencerkclark: Indeed, I think the order matters here, so we need to preserve that!

aospy/utils/vertcoord.py
spencerkclark: Yes, that's much nicer.

spencerkclark:
> But it feels a bit out of place there, like we're shimming it in. What do you think?

Yes, it did feel a shame to make `load_variable` more complicated.  I went with your suggestion in my latest commit: 9829549.

aospy/calc.py
```diff
@@ -13,35 +13,32 @@
...
+    the dtype_in_vert."""
+    vars_out = []
+    for var in variables:
+        if var.name == 'p':
+            vars_out.append(_P_VARS[dtype_in_vert])
+        elif var.name == 'dp':
+            vars_out.append(_DP_VARS[dtype_in_vert])
+        else:
+            vars_out.append(var)
+    return vars_out
```
spencerkclark: @spencerahill to aid in testing this I think we may want to implement a `__eq__` method in `Var`.  I need some way to verify that the correct `Var` object is substituted in.  What do you think?

spencerahill: I see what you mean.  Related to but not quite #170.

What would Var equality look like?  For starters it would seem that `name` and `func` should be equal.  Or at least that should be enough for the purposes of this test.  Right?  I'm inclined to leave a full `__eq__` implementation for another time, as it deserves more thought.

spencerahill:
>I went with your suggestion in my latest commit: 9829549.

Thanks.  Initially, I was envisioning moving it out of DataLoader entirely, but now I see that how you've done it is better...it enables DataLoaders of variables that rely on grid attrs to be entirely self-contained, rather than relying on Calc to grab the grid attrs for them if they can't be found on disk otherwise.

spencerkclark:
Indeed my primary motivation for doing that was to be able to support recursive calculations where a `Var` object standing in for a grid attribute was required for computing a variable that was not in the top level of recursion.

spencerkclark:
@spencerahill when you get a chance I think this should be ready for a final review.

aospy/test/test_calc_basic.py
```diff
@@ -275,5 +278,57 @@ def test_recursive_calculation(recursive_test_params):
...
+    _test_files_and_attrs(calc, 'av')
+
+
+@pytest.mark.parametrize(
+    ['dtype_in_vert', 'expected'],
+    [(ETA_STR, [p_eta, dp_eta, condensation_rain, 5]),
+     ('pressure', [p_level, dp_level, condensation_rain, 5])])
+def test_replace_pressure(dtype_in_vert, expected):
+    arguments = [p, dp, condensation_rain, 5]
+    p_in, dp_in, _, _ = arguments
```
spencerahill: Minor improvement: rather than discarding the non-pressure arguments, assert that they are identical before/after the call

aospy/test/test_data_loader.py
```diff
@@ -788,5 +788,43 @@ def test_recursively_compute_variable_multi_level(load_variable_data_loader):
...
+        DatetimeNoLeap(5, 12, 31), model=example_model,
+        intvl_in='monthly')
+    filepath = os.path.join(os.path.split(ROOT_PATH)[0], 'netcdf',
+                            '00060101.sphum_monthly.nc')
+    expected = 3 * _open_ds_catch_warnings(filepath)['bk']
+    np.testing.assert_array_equal(result.values, expected.values)
+
+
+def test_recursively_compute_grid_attr_error(load_variable_data_loader):
+    zsurf = Var(name=ZSURF_STR, def_time=False, def_vert=False,
```
spencerahill: Could you add a one-line comment explaining why this should fail?  Not self evident otherwise.

spencerahill (commented):
Two very minor comments, then please merge at your convenience once tests are green; or I will eventually.  Thanks!

spencerkclark:
Thanks for your comments @spencerahill!  Merging now.

spencerahill:
Thanks @spencerkclark!  Along with my smaller ones preceding this, we've really started to make a dent in cleaning up `calc.py`.  And all the more so once we get the Reductions in.

PR #288, opened 2018-06-16
Get rid of dask warning; always delete test output files

spencerahill:
`dask.set_options` -> `dask.config.set` as of dask version 0.18.  I've made this conditional on the dask version.  Alternatively we could just bump the minimum version to 0.18 so not have to deal with the version-checking logic.

Also, apologies for the 2-in-1, delete output of new tests from #287, as we do with other calc tests.


aospy/automate.py
```diff
@@ -278,7 +278,11 @@ def _compute_or_skip_on_error(calc, compute_kwargs):
 def _submit_calcs_on_client(calcs, client, func):
     """Submit calculations via dask.bag and a distributed client"""
     logging.info('Connected to client: {}'.format(client))
-    with dask.set_options(get=client.get):
+    if float(dask.__version__[:4]) < 0.18:
```
spencerkclark: I think xarray typically does things like this with `LooseVersion` from the standard library (e.g. [here](https://github.com/pydata/xarray/blob/cf19528d6d2baf988ad34e024cae28361c9fd693/xarray/tests/__init__.py#L91)):
```python
from distutils.version import LooseVersion

if LooseVersion(dask.__version__) < '0.18':
    dask_option_setter = dask.set_options
else:
    dask_option_setter = dask.config.set
```

spencerkclark (commented):
> dask.set_options -> dask.config.set as of dask version 0.18.

Very nice, I'm excited that that's been released.

> Also, apologies for the 2-in-1, delete output of new tests from #287, as we do with other calc tests.

Ah, good catch!  Thanks for cleaning that up.

spencerahill:
Thanks @spencerkclark.  I had seen their use of LooseVersion before but I had assumed it was a 3rd party package...that's great that it's standard library.

PR #289, opened 2018-06-18
WIP Custom time reductions

### LINKED ISSUE #208: Adding custom reduction methods

spencerahill:
Closes #208 

This is an initial stab at the implementation of the custom time reduction methods.

New since the most recent discussion in #208: I implemented a generic Reduction class and two subclasses, `GriddedReduction` and `RegionReduction`, in order to handle the fact the regional averaging of time-dependent vertical coordinates.

An outstanding questions: how to handle identifying time defined reductions.  For the built-in reductions, we can just specify them manually as we have been.  But this won't work for custom reductions.  And the way we're doing it now based on the string labels seems hackish to me.

Beyond that, I would appreciate an initial review @spencerkclark when you get a chance...does this seem like it's heading in the right direction?  I'm 100% receptive to changing any of it.  No rush...I'm basically out of spare cycles for the rest of this week.

Also still needed:

- [ ] Full test coverage
- [ ] Full documentation
- [ ] What's new entry


spencerahill:
Hmmm not sure what's up with the stickler comments...all tests are passing on my machine.

Also forgot to mention: the changes to Region aren't strictly needed for this PR, so I shouldn't have put them here.  But now that they are here: the motivation was that previously `Region.mask_var` only masked based on the lat/lon bounds, not the land mask; the land mask logic was handled in `ts`.  So I moved that from `ts` into `mask_var`, with the option to ignore it (which is physically necessary when masking the surface area).

aospy/calc.py
```diff
@@ -334,123 +371,123 @@ def _get_all_data(self, start_date, end_date):
...
+        # Load.
+        data_in = self._get_all_data(self.start_date, self.end_date)
+        # Compute.
+        data_out = self._function(*data_in).rename(self.name)
+        data_out = _time_weights_as_days(data_out)
+        # Reduce.
+        data_out = self._apply_vert_reduc(data_out)
+        gridded_out = self._apply_gridded_reductions(data_out)
+        region_out = self._apply_region_reductions(data_out)
+        combined_out = {**gridded_out, **region_out}
```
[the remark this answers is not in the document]
spencerkclark: > Hmmm not sure what's up with the stickler comments...all tests are passing on my machine.

I think this syntax requires Python 3.5 or above (this is fine since we have dropped Python 2 support).  Maybe modifying our `.stickler.yml` to look like the following could do the trick?
```
linters:
  flake8:
    python: 3
    fixer: true
```
See here for more information: https://stickler-ci.com/docs#python.  `python: 3` means Python 3.6 for Stickler.

spencerkclark:
Thanks @spencerahill!  I'll look have a look over this over the next few days.

aospy/calc.py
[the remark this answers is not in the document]
spencerahill: Good catch!  That did the trick.

aospy/calc.py
```diff
@@ -209,13 +225,34 @@ def __init__(self, proj=None, model=None, run=None, var=None,
...
+                time_reg_reducs.append(GriddedReduction.instances['full'])
+            else:
+                raise ValueError("Each element of `dtype_out_time` must be "
+                                 "a GriddedReduction object, RegionReduction "
+                                 "object, or a string corresponding to the "
+                                 "label of one of these reduction objects.  "
+                                 "Bad value given: {}".format(dtype))
+        self.dtype_out_time = tuple(time_reg_reducs)
+
+        if (self.dtype_in_time == 'av') or (not self.def_time):
```
spencerkclark, line 252: nit: the parentheses are not needed here

aospy/reductions.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+reg_no_reduc = RegionReduction(_null_reduc, 'full')
+reg_time_avg = RegionReduction(_time_avg, 'av')
+reg_yearly_ts = RegionReduction(_yearly_avg, 'ts')
+reg_yearly_stdev = RegionReduction(_yearly_stdev, 'std',
+                                   coord_reduction=reg_yearly_ts)
+
+
+_TIME_DEFINED_REDUCTIONS = ('av', 'ts', 'std', 'reg.av', 'reg.ts', 'reg.std')
```
spencerkclark, line 137: > An outstanding questions: how to handle identifying time defined reductions.  For the built-in reductions, we can just specify them manually as we have been. But this won't work for custom reductions.

I agree; I think a reasonably clean way of doing this would be to give `Reduction` objects an attribute that flags whether or not it reduces along time (maybe called something like `reduces_time`), which one could set in the constructor.  Then checking whether or not it was time-defined would be simply a manner of checking `if reduction.reduces_time: ...`

Another option would be to generalize things a bit more and let one specify a `dims` argument in the `Reduction` constructor, which would accept either a single string (which we could coerce to a list) or a list of strings.  Then the check might look like: `if TIME_STR in reduction.dims: ...`  This has the benefit that it might allow for future dimension-dependent logic in aospy (other than just with time).

Both options overcome the issue of how to deal with this for custom reductions.

aospy/reductions.py
```diff
@@ -0,0 +1,137 @@
...
+    def label(self, value):
+        raise ValueError("'label' method cannot be modified after Reduction "
+                         "object has been created.")
+
+    def __call__(self, data, *args, **kwargs):
+        return self.reduce(data, *args, **kwargs)
+
+
+class GriddedReduction(Reduction):
+    # TODO: I think I am overwriting the parent class's `instances` attribute.
```
spencerkclark, line 58: Is there anything against renaming `instances` something like `_REDUCTION_INSTANCES` and making it a global variable within this module?

aospy/reductions.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @label.setter
+    def label(self, value):
+        raise ValueError("'label' method cannot be modified after Reduction "
+                         "object has been created.")
+
+    def __call__(self, data, region, *args, coord_name=None, coord_data=None,
+                 **kwargs):
+        data_out = self.reduce(region.ts(data))
+        if self._coord_reducer is not None and coord_name is not None:
```
spencerkclark, line 100: Could it be more transparent here to have a single keyword argument called `coord` (default `None`) which would accept either a string or a DataArray?  Then the logic here could look like:
```python
if self._coord_reducer is not None and coord is not None:
    if isinstance(coord, str):
        coord_name = coord
        coord_data = data[coord_name]
    elif isinstance(coord, xr.DataArray):
        coord_name = coord.name
        coord_data = coord
    else:
        raise ValueError("'coord' must either be of type str or DataArray")
    data_out[coord_name] = self._coord_reducer(coord_data, *args, **kwargs)
```

aospy/calc.py
```diff
@@ -334,123 +371,123 @@ def _get_all_data(self, start_date, end_date):
...
+            Whether to write the results of the calculation to a tarball
+            archive.  Default True.
+
+        Returns
+        -------
+        This ``Calc`` object, whose ``data_out`` attribute has been created or
+        updated with the results of the calculations performed.
+
+        """
+        # Load.
```
spencerkclark, line 466: The new docstring is great!  I think we can remove these one word comments.

aospy/calc.py
```diff
@@ -294,10 +332,9 @@ def _add_grid_attributes(self, ds):
                 self.pressure = ds.level
         return ds
 
+    # @log_step(logging.info, "Getting input data: {}".format(self.var))
```
spencerkclark, line 335: Were calls like this generating name errors for `self.var`?

spencerkclark (commented):
This is looking pretty solid!  I left a couple suggestions inline for how to handle identifying time-defined reductions.  Outside of test coverage and documentation, I don't think there is too much else left to worry about for this PR; I'm fine leaving the region parts in if you want.

spencerahill:
Thanks @spencerkclark.  Glad it's on the right track.  Will have a bit of time next week to return to this, but frankly might be after that when I can get it over the finish line.

aospy/reductions.py
spencerahill: >Another option would be to generalize things a bit more and let one specify a dims argument in the Reduction constructor, which would accept either a single string (which we could coerce to a list) or a list of strings. Then the check might look like: if TIME_STR in reduction.dims: ... This has the benefit that it might allow for future dimension-dependent logic in aospy (other than just with time).

I like this.  In doing this PR it has occured to me that our current separation of reductions with respect to different dimensions is somewhat (albeit not entirely) arbitrary.

aospy/reductions.py
spencerahill: I hadn't thought of that.  But then if users define custom reductions elsewhere, will this still work?  Maybe yes; I haven't tried it yet.

aospy/reductions.py
spencerahill: Yes I like your idea.

aospy/calc.py
spencerahill: Basically yes.  I got a little too cute with trying to switch the logging to be all in terms of decorators (and which is out of scope for this PR anyways).  For those logging messages that rely on the method's arguments, we'll just have to retain the within-method logging call.

One could stay with the pure decorator approach by creating wrapper methods, but that seems to be more trouble than its worth.  I might just revert all of these logging changes, as they probably deserve their own consideration and PR.

aospy/reductions.py
spencerkclark: I'm not 100% sure, but I think it should work the same way.

aospy/reductions.py
spencerkclark: > In doing this PR it has occured to me that our current separation of reductions with respect to different dimensions is somewhat (albeit not entirely) arbitrary.

Agreed!  I think at some point it could be nice to consolidate things, but for now we can keep things as they are.

Issue #293, opened 2018-11-03
Data with coordinate variables (lat, lon, etc.) with names not included in internal_names


jdossgollin:
This may be two separate questions, but longitudes are often treated differently across models. Specifically:
1) Is there a way to define `alt_names` for a coordinate so that we can deal with something which is `X`, `lon`, or `longitude?
2) Is there a way to join longitudes that go from 0 to 360 with longitudes that go from -180 to 180 so that the outputs are standardized and everything works well with regions?
This seems like something which would go in the `model` creation

spencerkclark:
> 1. Is there a way to define alt_names for a coordinate so that we can deal with something which is `X`, `lon`, or `longitude`?

Indeed coordinates with similar but different names are so common that we have built a number of commonly-used internal names into aospy (in [internal_names.py](https://github.com/spencerahill/aospy/blob/develop/aospy/internal_names.py#L50-L70)): 

```
# All attributes associated with data's spatiotemporal grid.
GRID_ATTRS = OrderedDict(
    [(LAT_STR, ('lat', 'latitude', 'LATITUDE', 'y', 'yto', 'XLAT')),
     (LAT_BOUNDS_STR, ('latb', 'lat_bnds', 'lat_bounds')),
     (LON_STR, ('lon', 'longitude', 'LONGITUDE', 'x', 'xto', 'XLONG')),
     (LON_BOUNDS_STR, ('lonb', 'lon_bnds', 'lon_bounds')),
     (ZSURF_STR, ('zsurf', 'HGT')),
     (SFC_AREA_STR, ('area', 'sfc_area')),
     (LAND_MASK_STR, ('land_mask', 'LANDFRAC', 'XLAND')),
     (PK_STR, ('pk',)),
     (BK_STR, ('bk',)),
     (PHALF_STR, ('phalf',)),
     (PFULL_STR, ('pfull',)),
     (PLEVEL_STR, ('level', 'lev', 'plev')),
     (TIME_STR, ('time', 'XTIME')),
     (TIME_WEIGHTS_STR, ('time_weights', 'average_DT',)),
     (TIME_BOUNDS_STR, ('time_bounds', 'time_bnds')),
     (BOUNDS_STR, ('bounds', 'bnds', 'nv', 'nbnd')),
     (RAW_START_DATE_STR, ('raw_data_start_date',)),
     (RAW_END_DATE_STR, ('raw_data_end_date',))]
)
```

[...]

PR #296, opened 2018-11-04
Modifications to follow the IRIDL conventions

jdossgollin:
This PR would make the `GRIDD_ATTRS` more compatible with the IRIDL conventions, based on the Columbia IRI data library (http://iridl.ldeo.columbia.edu/). IRIDL conventions include `X` for longitude, `Y` for latitude, and `T` for time, which are the three changes made in this PR.

See http://iridl.ldeo.columbia.edu/SOURCES/.NOAA/.NCEP-NCAR/.CDAS-1/.DAILY/.Intrinsic/.PressureLevel/.temp/.

I would like to include `P` for pressure level for variables that use pressure coordinates, but this modification seems non-trivial.


spencerkclark:
This looks great, thanks!  The build failures (i.e. Appveyor and Travis) are all unrelated.  To get things to pass for now, go ahead and add an `xfail` decorator to `test_load_variable_does_not_warn` in `test_data_loader.py`.  For example:
```
@pytest.mark.xfail
@pytest.mark.parametrize(['start_date', 'end_date'],
                         _LOAD_VAR_DATE_RANGES.values(),
                         ids=list(_LOAD_VAR_DATE_RANGES.keys()))
def test_load_variable_does_not_warn(load_variable_data_loader,
                                     start_date, end_date):
```
With the latest released versions of pandas and xarray, there is currently an unrelated warning raised there:
```
aospy/test/test_data_loader.py::test_load_variable_does_not_warn[cftime]
  //anaconda/envs/aospy_dev/lib/python3.6/site-packages/xarray/coding/cftimeindex.py:160: FutureWarning: CFTimeIndex.data is deprecated and will be removed in a future version
    if self.data:
```
I fixed this upstream in xarray (pydata/xarray#2448), so it should no longer emit a warning once xarray version 0.11 is released, which should be soon.

aospy/internal_names.py
```diff
@@ -61,7 +61,7 @@
      (PHALF_STR, ('phalf',)),
      (PFULL_STR, ('pfull',)),
      (PLEVEL_STR, ('level', 'lev', 'plev')),
-     (TIME_STR, ('time', 'XTIME')),
+     (TIME_STR, ('time', 'XTIME', 'T')),
```
spencerkclark, line 64: I'll defer to @spencerahill's judgement on this one, but I'll note that `'T'` is also used to represent temperature in the WRF model, so we might hold off on adding this as an automatic alternative name for `'time'`.  I've fast-tracked an implementation of #182 (see #297), which might be the preferred solution in this particular case.

spencerahill: Thanks @spencerkclark, I agree that with #297 now in the pipeline, let's limit this PR to 'X' and 'Y', since we can be pretty sure that those won't cause any problems in the way that 'T' might.

@jdossgollin, does that work for you?  Thanks much for the PR!

jdossgollin: Yes works for me -- in discussion of #293 I suggested that a `rename` in the `preprocess` function could also take care of something like time as `T`.

I'm sorry to say I'm not sure what I should do to finish patching this PR through -- do I edit the code in here and commit back in? Thanks for the patience!

spencerahill: On your local machine, make sure you're still on this branch, make your edits, `git commit` them, and then once you're ready `git push` the branch again to here.  The PR will be updated accordingly.

Please also add an entry to our [What's New](https://github.com/spencerahill/aospy/blob/develop/docs/whats-new.rst) docs page.  Just mimic the structure of the other entries, under the v0.3 heading (don't worry about the v0.3 vs. v0.2.1 bit; we'll fix that later).

Finally, for almost all changes or additions to the codebase, we require formal tests that ensure that the new code actually works as we intended and didn't break anything unintentionally.  These are in the [test](https://github.com/spencerahill/aospy/tree/develop/aospy/test) directory.  However, this may be one of the rare exception where we're OK without.  I don't think we have any existing tests specifically of the different options in `internal_names`, so it would be a lot to ask of you to create it.  @spencerkclark what do you think?

spencerkclark: I agree; I think we're OK without tests in this case.  The grid attribute renaming method is exercised by the tests [here](https://github.com/spencerahill/aospy/blob/a2cd51fc1cb33c0667e02b6d3a14ac529784f553/aospy/test/test_data_loader.py#L155-L192).  If things work for one alternate name/internal name pair (in the case of the tests `'LATITUDE'`->`'lat'`) then I would expect things should work for another pair just as well.

spencerkclark: @jdossgollin a nice (fairly comprehensive) resource for learning about Git/GitHub/testing/documentation in the context of contributing to open source Python projects is the [xarray contributing guide](http://xarray.pydata.org/en/stable/contributing.html).  In an abstract sense (i.e. replace xarray with aospy) much of what is in there applies here.

spencerahill:
C.f. https://github.com/spencerahill/aospy/issues/293#issuecomment-437542424, @jdossgollin I'd say all that's left here is adding `P` for pressure.  And a What's New entry.

Edit: and `nbnds` for `BOUNDS_STR`, c.f. https://github.com/spencerahill/aospy/issues/299#issuecomment-437084968

jdossgollin:
Note to self: see also discussion in #293 comments

spencerkclark:
Also a quick note, what I wrote here, https://github.com/spencerahill/aospy/pull/296#issuecomment-435674393, is no longer needed (since xarray 0.11 is now out).

jdossgollin:
Note comment at https://github.com/spencerahill/aospy/pull/296#issuecomment-437548913

PR #297, opened 2018-11-04
Add grid_attrs argument to Model constructor

### LINKED ISSUE #182: Support user-specified internal names of grid attributes

spencerkclark:
Closes #182


spencerahill:
Thanks for speeding this through @spencerkclark!  I should have time on Friday to properly review this.

aospy/test/data/objects/examples.py
```diff
@@ -39,7 +41,8 @@ def total_precipitation(convection_rain, condensation_rain):
                       'im.landmask.nc')),
     ),
     runs=[example_run],
-    load_grid_data=True
+    load_grid_data=True,
+    grid_attrs={LAND_MASK_STR: 'custom_land_mask', LON_STR: 'custom_lon'}
```
spencerahill, line 45: Nice, I think this is a good idea to have modified the example data directly to ensure that this works and be able to show this functionality in the tutorial.

docs/examples.rst
```diff
@@ -158,16 +158,28 @@ The other important attribute is ``runs``, which is a list of the
 :py:class:`aospy.Run` objects that pertain to simulations performed in
 this particular model.
 
+Lastly, grid attibutes (like time, latitude, longitude, etc.), often have
+different names from model to model.  ``aospy`` can handle most typical grid
+attribute names automatically; however, if your model uses unusual grid
+attribute names you can add a ``grid_attrs`` dictionary to your
+:py:class:`~aospy.Model` object to define a mapping between ``aospy`` internal
+names and names for the grid attributes that can be found in your data files.
+As an example, here only the land mask and longitude variables have atypical
```
spencerahill: I think it would help to be even more explicit, e.g. something like "In our example data, the land mask and longitude variables both have non-standard names: 'custom_land_mask' and 'custom_lon', respectively....

spencerahill (commented):
With the caveat that I'm a little rusty at thinking carefully about the codebase, I actually think the implementation looks good as is.  A couple minor docs suggestions is all.

docs/examples.rst
spencerkclark: Good call.  I updated things in my latest push.

spencerahill:
Awesome --- this looks ready to go @spencerkclark .  Merge whenever you're ready, and thanks as always for the great work.

spencerkclark:
@spencerahill one more minor update for you to have a look at; I added a reference table to the documentation showing the current built-in alternative names (I think it's a bit tricky for new users to find those at the moment).  Let me know if that looks good and I'll go ahead and merge.

docs/examples.rst
```diff
@@ -158,18 +158,79 @@ The other important attribute is ``runs``, which is a list of the
...
+.. _built-in-alternative-names:
+
+Built in alternative names
+**************************
+
+For reference, here is a table showing the list of grid attributes used in
+``aospy`` along with their associated built-in alternative names.  We are
+always open to adding more alternative names to this list, as long as their
+association with particular grid attributes is sufficiently unambiguous.
+For potentially ambiguous alternative names, e.g. 'T' for ``TIME_STR``, we
```
spencerahill: One tiny edit: I would expand on why 'T' is potentially ambiguous, e.g. something like "... e.g. 'T' for ``TIME_STR`` (since 'T' is used for time in some datasets but for temperature in others) ..."

spencerahill (commented):
Great idea on the table.  One tiny edit; go ahead and merge once that's done.

docs/examples.rst
spencerkclark: Thanks, that's a good addition.

spencerkclark:
Thanks for the review @spencerahill; merging now!

PR #304, opened 2018-11-09
Minor fixes to suppress warnings in xarray 0.11

spencerkclark:
This silences a few deprecation warnings in loading variables and in our test suite that came up with the latest update to xarray


spencerahill:
Thanks @spencerkclark .  Any idea what the cause is of the 4 failures for 3.5?

spencerkclark:
It seems that conda-forge is no longer building packages for Python 3.5 (and has moved on to 3.6 and 3.7), so our 3.5 environment is using xarray 0.10.9 (which was the last version of xarray built for that platform).  When I get a chance I'll configure things to ignore that particular warning (which is causing the test to fail).

The notebook test on Windows is a bit more puzzling.  I'll have to think about that one more.

spencerahill:
OK, thanks for sorting that out.  I suppose we can have the conversation of whether to keep supporting 3.5 or keep up with conda.  FWIW I'm fine with merging this as is and dealing with that separately.  Although yes I suppose we should figure out the AppVeyor thing first.

spencerahill:
Yikes, AppVeyor thing looks not fun.  I don't want you to kill a bunch of time on that, which is really not important.

Whether or not you get that fixed, go ahead and merge when you're ready.  If it still needs to be fixed, just open a new issue to track it.

spencerkclark:
> Yikes, AppVeyor thing looks not fun. I don't want you to kill a bunch of time on that, which is really not important.

Yeah let's put it off for now...it's odd that it would stop working all of a sudden there, but it's hard to diagnose the issue without being able to test things locally.  I [borrowed some logic from xarray](https://github.com/pydata/xarray/blob/575e97aef405c9b473508f5bc0e66332df4930f3/xarray/tests/test_backends.py#L58) to detect whether we are on Windows or not, so hopefully the test should be skipped now by AppVeyor.

PR #306, opened 2018-11-12
Add support for Python 3.7 and corresponding CI envs

### LINKED ISSUE #292: Add Python 3.7 support
Should be straightforward, c.f. https://github.com/pydata/xarray/pull/2271

Although we might as well wait until they've merged that one, including figuring out the Travis issues referenced in that PR.

spencerahill:
Closes #292.  

Also switched Travis language to 'generic', c.f. https://github.com/pydata/xarray/pull/2432


spencerahill:
Rare case where everything seemed to work on the first try.  @spencerkclark any suggestions/reservations on this before merging?

spencerkclark (commented):
Looks great; I'm glad this went so smoothly!  Thanks for taking care of it.  A couple minor suggestions while we're updating the CI setup:
- Could you add a `conda list` call directly after we activate the test environment under the `install` header in the Travis and AppVeyor yml files?  I find this facilitates determining the versions of packages installed in the environment, which is useful for debugging purposes.  Currently I use the labels associated with each package's download, but those are truncated (so are sometimes missing information) and are also not in alphabetical order (which makes it slow to find the package I'm looking for).
- Since the AppVeyor builds run serially (i.e. slowly), maybe we should consider dropping the Python 3.5 build there?

spencerkclark (approved):

spencerahill:
Good suggestions @spencerkclark .  In it goes!

Issue #308, opened 2018-11-12
Use `stable` and `master` as our two main branches, rather than `master` and `develop`


spencerahill:
C.f. https://github.com/spencerahill/aospy/pull/305#issuecomment-438035813.  Makes us more in-sync with xarray and reduces confusion for new contributors.

spencerahill:
Done.

Issue #307, opened 2018-11-12
Towards v0.3 release


spencerahill:
Although I haven't finished #289, I think we now have more than enough new features and breaking changes to merit a v0.3 release.  I can turn around a quick v0.3.1 release once I finish #289.

Realistically I can probably get to this not this week but sometime in the following couple weeks.  But I'd say it's worth waiting on the pending PRs by @jdossgollin.  @jdossgollin, think you can manage to wrap those up in the next week or so?

jdossgollin:
Yes -- tonight or tomorrow (knock on wood 😄)

spencerahill:
In doing this release, I will also fix #308

spencerkclark:
It looks like we'll need to update the pinned versions in `docs/environment.yml` for the documentation to build:
https://readthedocs.org/projects/aospy/builds/8102985/

spencerahill:
>It looks like we'll need to update the pinned versions in docs/environment.yml for the documentation to build

Yes, just realized that.  Have pushed commits that hopefully fix that.

spencerahill:
Those commits are basically copying https://github.com/pydata/xarray/pull/2310

I'm just committing them to master directly rather than as PRs since this is all just to fix the docs

spencerahill:
The v0.3.0 release is done on github and on pip.  I'm having a little trouble getting our feedstock updated for the release on conda; waiting for help from the conda-forge folks on that.

For the docs, there are the following problems: 
- the 'stable' and 'latest' builds are completing, but 'stable' is still only showing v0.2 rather than v0.3
- the 'v0.3.0' build is failing, because of the incorrect settings in docs/environment.yml noted above.  I'm not sure how to fix that, since the release has already been made.

Once we resolve all these things and we're up on conda-forge, I'll make the release email announcements

spencerkclark:
> the 'v0.3.0' build is failing, because of the incorrect settings in docs/environment.yml noted above. I'm not sure how to fix that, since the release has already been made.

Hmm...this one is tricky.  Is it considered bad practice to [delete a release](https://help.github.com/articles/editing-and-deleting-releases/) and redo things?

spencerahill:
>Is it considered bad practice to delete a release and redo things?

Good thinking.  That may be the only way to go.  I'll ask the RTD people first to see if there's a less intrusive option.

---

[...]

PR #311, opened 2018-11-14
Use Versioneer and add HOW_TO_RELEASE

spencerahill:
In anticipation of the v0.3 release (#307), this PR does two things (arguably more appropriately two separate PRs, but since they both relate to releasing I figured it's fair):

1. C.f. https://github.com/pydata/xarray/pull/2163, use Versioneer to simplify keeping track/updating version number.  I basically followed how xarray did it exactly, including copying in the versioneer code directly into our codebase (Versioneer is public domain, so there's no worry re: licensing) rather than making it a dependency.

2. Add a HOW_TO_RELEASE document, again copied nearly verbatim from xarray's.  NOTE: this includes the use of stable and master branches, rather than our current use of master and develop.  But since we will be making that switch when we bump to v0.3 anyways (#308), and this document is mostly for my own reference anyways, I figure it's OK for it to leave it as is for now.


spencerkclark (commented):
Very nice!  Judging by the Coveralls report I think we may want to specify that we want to omit `aospy/_version.py` in our `.coveragerc` file.

spencerahill:
Good call @spencerkclark.  Any other edits?  Otherwise will merge this later today.

spencerkclark (approved):
Thanks @spencerahill this looks good to me.

PR #312, opened 2018-11-14
v0.3.0 release

spencerahill:
Trying to do the master/develop -> stable/master switch at the same time (#308), so this might have some hiccups.


spencerahill:
Actually don't need this, which would merge the new changes into `develop`.  The whole point is that we're getting rid of `develop`.  So closing this.

PR #313, opened 2018-11-16
Only test on 3.7 for Appveyor

spencerahill:
Appveyor is slow, and we have no known windows users yet, so I think we can get away with only testing against 3.7.  We can revisit if/when problems on windows start to arise.


spencerahill:
The one Travis failure was from the flaky distributed tests, so we're good.  Merging.

spencerkclark:
👍 I agree.

PR #315, opened 2018-11-19
Add missing ipython:: python directive to fix docs builds with latest…

spencerahill:
… versions of sphinx

I had updated sphinx on my local machine to 1.8.2 and came across this.  One of our ipython blocks was missing the `python` after `.. ipython:: `, which apparently in recent sphinx versions results in an error.


spencerkclark (approved):
Good catch!

Issue #320, opened 2019-03-30
New failure in test_apply_time_offset


spencerahill:
After reverting to xarray 0.11, c.f. #319, the test suite loads without error but has one failure, which I've pasted below.  @spencerkclark, since this seems cftime-related (or adjacent), I'm going to punt this one to you.

```python
test_apply_time_offset ___________________________________________________________________

cls = <class 'pandas.core.arrays.datetimes.DatetimeArray'>
index = <DatetimeArray>
['1898-06-30 00:00:00', '1898-07-30 00:00:00']
Length: 2, dtype: datetime64[ns], freq = <MonthEnd>, kwargs = {}, inferred = None
on_freq = <DatetimeArray>
['1898-06-30 00:00:00', '1898-07-31 00:00:00']
Length: 2, dtype: datetime64[ns]

    @classmethod
    def _validate_frequency(cls, index, freq, **kwargs):
        """
        Validate that a frequency is compatible with the values of a given
        Datetime Array/Index or Timedelta Array/Index

        Parameters
        ----------
        index : DatetimeIndex or TimedeltaIndex
            The index on which to determine if the given frequency is valid
        freq : DateOffset
            The frequency to validate
        """
        if is_period_dtype(cls):
            # Frequency validation is not meaningful for Period Array/Index
            return None

        inferred = index.inferred_freq
        if index.size == 0 or inferred == freq.freqstr:
            return None

        try:
            on_freq = cls._generate_range(start=index[0], end=None,
                                          periods=len(index), freq=freq,
                                          **kwargs)
            if not np.array_equal(index.asi8, on_freq.asi8):
>               raise ValueError
E               ValueError

../../pandas/core/arrays/datetimelike.py:884: ValueError

During handling of the above exception, another exception occurred:

    def test_apply_time_offset():
        start = datetime.datetime(1900, 5, 10)
        years, months, days, hours = -2, 1, 7, 3
        # test lengths 0, 1, and >1 of input time array
        for periods in range(3):
            times = pd.date_range(start=start, freq='M', periods=periods)
            actual = apply_time_offset(xr.DataArray(times), years=years,
                                       months=months, days=days, hours=hours)
            desired = (times + pd.tseries.offsets.DateOffset(
>               years=years, months=months, days=days, hours=hours
            ))

test_utils_times.py:55:
[...]

PR #321, opened 2019-03-30
WIP fixes for recent CI build failures

### LINKED ISSUE #319: Failing tests in CI, but for some builds still come back as green
I recently tried running our test suite on a freshly created conda environment for aospy development, and I got some weird errors.  To see if they were just a local problem, I re-triggered the Travis builds on our master branch, and they appeared [there](https://travis-ci.org/spencerahill/aospy/builds/457237155) too.  There are two different ones:

1. On our [docs env build](https://travis-ci.org/spencerahill/aospy/jobs/457237159), which causes the build to register as a failure.
2. On all the other builds (e.g. [py37](https://travis-ci.org/spencerahill/aospy/jobs/457237158)), which despite the errors that prevent the test suite from even loading, the build registers as a success.  So I suspect that because the tests never even get ran, they can't generate a failure, even though obviously this should be a failure.

So there are three things to fix:

1. Fix the bug in the docs branch.  The error involves an import within dask, and unlike our other builds with dask=1.15.2, the dask version in the docs build is 0.18.2.
2. Fix the bug in the other branches.  I'm pretty sure this stems from https://github.com/pydata/xarray/pull/2782 which made it into xarray v0.12, as when I revert xarray to 0.11, the tests do load on my local machine.
3. Fix our CI build process so that a failure to load the tests constitutes a failing build.

spencerahill:
Closes #319 

The new logic introduced in xarray v0.12 assumes the argument passed to `xr.open_dataset` includes a `tell` method.  In some cases when we are populating the grid data for `Model` objects, we pass it a tuple of paths.  Including `AttributeError` as one of the caught exceptions seems to do the trick.

I don't think this needs any new tests, since our existing test suite caught it (albeit crudely).

This addresses point 2 of #319.  The other two points remain.


spencerahill:
1a52855 addresses Point 1 of #319 (assuming it works).  So only point 3 remains.

spencerahill:
For point 3, i.e. that Travis was marking failing builds as passing, I'm pretty sure that the problem was that both our pytest command and the flake8 command were under the same `-` bullet point.  So even though the `pytest` command returns nonzero, the next line is the flake8 command, which returns zero, and so the whole item is interpreted as a pass.  So, copying xarray, I created a separate "lint" env for the flake8 check.

spencerahill:
All three points from #319 have now been fixed.  As such, paradoxically, I am happy to see failing tests, because those are from #320 and are at least being caught as failures!

So once we have a fix for #320 we should be back to passing tests *AND* for the right reasons this time.

spencerkclark (approved):
> All three points from #319 have now been fixed. As such, paradoxically, I am happy to see failing tests, because those are from #320 and are at least being caught as failures!

Hah!  Thanks for taking care of this @spencerahill.  This looks good to me.

PR #328, opened 2019-10-02
Turn on Azure Pipelines CI

spencerahill:
Closes #325 

Likely will take a few tries to get this working.

PR #329, opened 2019-10-02
Set up CI with Azure Pipelines

spencerahill:


spencerahill:
Just needed this to "turn on" the builds on the Azure side; rest of this will be handled by #328.

